Issue #4, opened 2025-06-30
Catch exceptions in the supplied log density function


WardBrian:
Right now, if the user-supplied log density throws an exception, the entire algorithm comes crashing down. 

We should probably turn these into divergences, either by returning `-inf` or through some other mechanism. It might also be nice
to generate a warning, though this requires us to include some idea of logging

PR #7, opened 2025-06-30
Update Stan example to compile, return -inf on failure

WardBrian:
Copies over some changes from test.cpp that got missed, and changes the test code to work around the lack of exception catching in the algorithm (#4).

This now means it can be used on relatively 'real' models like the SIR ode model. Before these would crash during initialization due to the errors Stan was throwing


bob-carpenter (approved):
Thanks.  Looks good

PR #8, opened 2025-07-11
Uturn ends

bob-carpenter:
This fixes a bug in the U-turn condition, which was previously defined with line

```cpp
  auto scaled_diff =
      (inv_mass.array() * (span_fw.theta_fw_ - span_fw.theta_bk_).array())
          .matrix();
```

The second argument to the difference should use `span_bk` rather than `span_fw`.

```cpp
  auto scaled_diff =
      (inv_mass.array() * (span_fw.theta_fw_ - span_bk.theta_bk_).array())
          .matrix();
```

had swapped `span_bk` for `span_fw`

### DIFF
--- a/tests/util_test.cpp
+++ b/tests/util_test.cpp
@@ -24,10 +24,6 @@ static Mat mat(S x00, S x01, S x10, S x11) {
   return y;
 }
 
-TEST(Util, Dummy) {
-  EXPECT_EQ(1 + 1, 2);
-}
-
 TEST(Util, Walnuts) {
   EXPECT_EQ(2 + 2, 4);
   Vec thetabk1 = vec(-3, 0);


tests/mock_test.cpp
WardBrian, line 1: (We might as well delete this now that we have a couple real tests)

tests/util_test.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+static Mat mat(S x00, S x01, S x10, S x11) {
+  Mat y(2, 2);
+  y << x00, x01, x10, x11;
+  return y;
+}
+
+TEST(Util, Dummy) {
+  EXPECT_EQ(1 + 1, 2);
+}
```
WardBrian, line 29:
```suggestion
```

WardBrian (approved):
Looks good, a couple nits.

As an aside: `dual_average_test` still fails on my machine. It would be nice to make the test more robust, so we could actually set up a small CI job

bob-carpenter:
Let's work on getting that test passing on your machine next week.  I don't know why its behavior would differ so much it would fail a pretty light test, but I doubt I'll be able to debug without access to your machine.

PR #9, opened 2025-07-14
filter logp_grad exceptions

bob-carpenter:
I added a wrapper around  logp and gradient functions to catch exceptions and return -infinity log density and zero gradients.  This seems easier than adding try/catch throughout the code.  Two things to note:

* The downside is that there's no way to play it unsafe by passing in an unwrapped function.

* The only tricky part is not doubling the wrapper, so I left the base log density function public so that `AdaptiveWalnuts` can pass the base log density function to the adapted `Walnuts` sampler.

### DIFF
--- a/include/walnuts/util.hpp
+++ b/include/walnuts/util.hpp
@@ -267,6 +267,7 @@ class NoExceptLogpGrad {
     try {
       logp_grad_(x, logp, grad);
     } catch (...) {
+      // TODO: log exception
       logp = -std::numeric_limits<S>::infinity();
       grad = Vec<S>::Zero(x.size());
     }


include/walnuts/util.hpp
```diff
@@ -236,4 +236,43 @@ inline bool uturn(const U &span1, const U &span2, const Vec<S> &inv_mass) {
...
+   * @param[out] logp The log density to set.
+   * @param[out] grad The gradient to set.
+   */
+  inline void operator()(const Vec<S>& x, S& logp, Vec<S>& grad)
+    const noexcept {
+    try {
+      logp_grad_(x, logp, grad);
+    } catch (...) {
+      logp = -std::numeric_limits<S>::infinity();
+      grad = Vec<S>::Zero(x.size());
```
WardBrian, line 271: I'd just add

```suggestion
	  // TODO: logging
      logp = -std::numeric_limits<S>::infinity();
      grad = Vec<S>::Zero(x.size());
```

Otherwise, I think this is good

PR #10, opened 2025-07-14
Add CI using github actions

WardBrian:
Apologies for the messy git history. I was testing this on https://github.com/WardBrian/walnuts/tree/main (github only lets you run actions that have at least once been in the default branch), and when I tried to push it over here it pushed to _our_ main as well. I've reverted that, so this PR is just a revert-of-the-revert

---

This adds a basic github action with four configurations:
- gcc on ubuntu
- clang on MacOS
- gcc on Windows
- MSVC on Windows

For each, it builds, runs the example program, and runs the tests. To get the tests passing on all four configurations I needed to mess with the seed and tolerance of the dual averaging test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,64 @@
+name: walnuts tests
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+  workflow_dispatch: {}
+
+env:
+  CACHE_VERSION: 1
+
+# only run one copy per PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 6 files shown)


bob-carpenter (approved):
I don't know anything about this continuous integration config.  Is @SteveBronder going to review that?  

The other changes are OK, but I liked `Integer` to control the type.  Should I just get rid of all of the `Integer` types and just replace with `auto` or `long`?

WardBrian:
Also we can ask Steve to review the actual config if you want, but if you just look at [the outputs](https://github.com/flatironinstitute/walnuts/actions/runs/16272017882?pr=10) that would probably be sufficient review

PR #12, opened 2025-07-15
Edits to readme

WardBrian:
The current readme contains some errors (e.g., it specifically mentions `ExternalProject_Add`, but we don't use that) and I think generally over-explains cmake. This is my attempt to keep the important bits while simplifying what we need to keep up to date over time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,214 +44,103 @@ information on its dependencies.
 * [BridgeStan](https://github.com/roualdes/bridgestan)  ([BSD-3
 licensed](https://opensource.org/license/bsd-3-clause))
 
+## Using WALNUTS in a C++ project
 
-## CMake Preliminary Build
+This library is header only, and only requires Eigen (also header only)
+for downstream usage. If your project uses CMake, you can depend on our
+`walnuts` library target. If not, any method of adding the `include/`
+folder of this repository to your build system's include paths should suffice
+as long as you also provide Eigen yourself.
 
-All build steps use CMake. The root configuration file is:
+## Building the examples and tests
 
-* `CMakeLists.txt`
+CMake is required to build the examples and tests.
 
-All instructions assume a `bash` shell (typically at `/usr/bin/bash`)
-and that commands are run from the top-level `walnuts` directory of
-the repository.
+### Configuring the build
 
-### Internet Connection Required
+The basic configuration is
 
-The build uses CMake's `FetchContent` and `ExternalProject_Add` to
-manage dependencies. **An internet connection is required when first
-building the project.**
-
-### Initial Configuration
-
-**Before building any targets**, you must configure the build with:
-
-```bash
-cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
+```sh
+# run from a directory you want the build results to be placed in
+# NOT required to be in the repository folder
+cmake <options> <repo_root>
 ```
 
-The options used are as follows.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,214 +44,102 @@ information on its dependencies.
...
 
-All build steps use CMake. The root configuration file is:
+## Building the examples and tests
 
-* `CMakeLists.txt`
+CMake is required to build the examples and tests.
 
-All instructions assume a `bash` shell (typically at `/usr/bin/bash`)
-and that commands are run from the top-level `walnuts` directory of
-the repository.
```
bob-carpenter, line 56: I'm often very confused in new projects about where I should be running commands from.  So something should say this somewhere.

README.md
```diff
@@ -44,214 +44,102 @@ information on its dependencies.
...
-building the project.**
-
-### Initial Configuration
-
-**Before building any targets**, you must configure the build with:
-
-```bash
-cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
+```sh
+cmake <options> <repo_root>
```
bob-carpenter, line 64: This would have confused me before.  I find it the most awkward part of Cmake compared to make other than the incredible verbosity.

README.md
```diff
@@ -44,214 +44,102 @@ information on its dependencies.
...
 ```
 
-The options used are as follows.
+where `<options>` are the CMake options and `<repo_root>` is the root
+directory of the repository (where `CMakeLists.txt` is found).
```
bob-carpenter, line 68: This is confusing---we have a plethora of `CMakeLists.txt` files.  You mean the top-level one I presume?

bob-carpenter (changes requested):
I'm OK writing this for devs, but there are still things in here I think we should point out that I signaled in comments.

README.md
WardBrian: We only have two by my count. I did mean the top level one here.

bob-carpenter (approved):
OK.  Not how I'd write it, but I'm willing to defer to you on whatever you want this to look like, including removing more of the instructions for beginners---I can just ask you when I get stuck.

README.md
```diff
@@ -44,214 +44,102 @@ information on its dependencies.
...
-```bash
-# /usr/bin/bash
-
-# remove old build, rebuild with --fresh to force a hard reset of cached variables
-rm -rf ./build
-cmake -S . -B "build" --fresh
-
-# All the cmake targets now exist in build
-cd build
-```
```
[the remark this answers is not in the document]
bob-carpenter: Too late.  Already merged.  I'm going to let you and @WardBrian hash this out as I really have zero opinion if this isn't being written for me.

README.md
```diff
@@ -44,214 +44,102 @@ information on its dependencies.
...
-```bash
-# /usr/bin/bash
-# Same as other command but -LH lists all cached cmake variables
-# along with their help comment
-cmake -S . -B "build" -LH | grep "WALNUTS" -B1
-
-# Output
-$ // Build the example targets for the library
-$ WALNUTS_BUILD_EXAMPLES:BOOL=ON
-```
```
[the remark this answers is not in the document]
bob-carpenter: Take it up with @WardBrian.  I'm sympathetic with not trying to teach CMake, but I have no idea where he's drawing the line or what an ordinary C++ dev can be expected to know about CMake.

PR #14, opened 2025-07-16
access tuning parameters

bob-carpenter:
I added methods on all the samplers to allow them to access the current mass matrix and step size and number of gradients called.  

I have other code for plotting all the output from the C++ tests, but that's not clean enough to commit yet.  I did instrument the tests to dump out the adapted mass matrix and step sizes.

### DIFF
--- a/examples/test.cpp
+++ b/examples/test.cpp
@@ -117,7 +117,7 @@ static void test_walnuts(const F& target_logp_grad, VectorS theta_init,
             << ";  max_error = " << max_error << std::endl;
   global_start_timer();
   nuts::Random<double, RNG> rand(rng);
-  nuts::WalnutsSampler sample(rand, target_logp_grad, theta_init, inv_mass,
+  nuts::WalnutsSampler<F, S, RNG> sample(rand, target_logp_grad, theta_init, inv_mass,
                               macro_step_size, max_nuts_depth, max_error);
   MatrixS draws(D, N);
   for (Integer n = 0; n < N; ++n) {
@@ -133,9 +133,9 @@ static void test_adaptive_walnuts(const F& target_logp_grad,
                                   Integer D, Integer N, Integer max_nuts_depth,
                                   S max_error) {
   Eigen::VectorXd mass_init = Eigen::VectorXd::Ones(D);
-  double init_count = 10.0;
-  double mass_iteration_offset = 4.0;
-  double additive_smoothing = 0.05;
+  double init_count = 2;
+  double mass_iteration_offset = 2;
+  double additive_smoothing = 0.001;
   nuts::MassAdaptConfig mass_cfg(mass_init, init_count, mass_iteration_offset,
                                  additive_smoothing);
   double step_size_init = 1.0;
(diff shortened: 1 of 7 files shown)


include/walnuts/util.hpp
```diff
@@ -273,7 +273,17 @@ class NoExceptLogpGrad {
...
+   * @brief Return the number of log density/gradient calls.
+   *
+   * @return The number of calls.
+   */
+  Integer logp_grad_calls() const noexcept {
+    return calls_;
+  }
+  
+  const F& logp_grad_;
+  Integer calls_;
```
WardBrian, line 286: I don’t think we necessarily should track this in this object. It seems simple enough for users who want it to do themselves, and it’s slightly annoying to have to remove all those const qualifiers as a result

examples/test.cpp
```diff
@@ -156,9 +156,31 @@ static void test_adaptive_walnuts(const F& target_logp_grad,
   global_start_timer();
   nuts::AdaptiveWalnuts walnuts(rng, target_logp_grad, theta_init, mass_cfg,
                                 step_cfg, walnuts_cfg);
+
+  std::ofstream out("adapt_mass.csv");
```
WardBrian, line 160: I feel like these changes to the example shouldn’t be committed to the main branch necessarily

WardBrian (changes requested):
I’d be happy to merge if this was just the inv_mass and step_size functions in AdaptiveWalnuts, but I’m not immediately on board the other changes in this patch (see specifics below)

include/walnuts/adaptive_walnuts.hpp
```diff
@@ -504,10 +505,39 @@ class AdaptiveWalnuts {
    * @return The WALNUTS sampler with current tuning parameter estimates.
    */
   WalnutsSampler<F, S, RNG> sampler() {
-    return WalnutsSampler<F, S, RNG>(
-        rand_, logp_grad_.logp_grad_, theta_, mass_estimator_.inv_mass_estimate(),
-        step_adapt_handler_.step_size(), walnuts_cfg_.max_nuts_depth_,
-        walnuts_cfg_.log_max_error_);
+    return WalnutsSampler<F, S, RNG>(rand_, logp_grad_, theta_,
+				     mass_estimator_.inv_mass_estimate(),
```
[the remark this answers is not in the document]
bob-carpenter: That was intentional.  Now it passes the wrapper and uses the copy constructor so that the count of gradients is preserved.  That's why there's now a new template parameter for that argument---it can be any old log density/gradient function or it can be a wrapped one.  I mention that behavior in the API doc, because it's a bit unusual, even though I don't like repeating things in the doc that are part of the C++ language spec.

include/walnuts/adaptive_walnuts.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <fstream>
```
WardBrian, line 3:
```suggestion

```

include/walnuts/walnuts.hpp
```diff
@@ -497,17 +496,22 @@ class WalnutsSampler {
...
-   * @param rand The randomizer for HMC.
-   * @param logp_grad The target log density and gradient function (see the
-   * class documentation.
+   * The template parameter `U` allows this function to either wrap a
+   * bare `logp_grad` function or accept a `LogpGrad` instance to
+   * copy (it will not double wrap).
+   *
+   * @tparam U The type of the target log density/gradient function.
+   * @param rand The compound randomizer for HMC.
+   * @param logp_grad The target log density/gradient function.
```
WardBrian, line 505: If you really want to preserve the log density count between the two, I think a more intuitive way to do this is to define a second constructor with the signature
`WalnutsSampler(Random<S, RNG> &rand, const LogpGrad<F> &logp_grad, const Vec<S> &theta, ... `

I think this would let you recover CTAD, at least

include/walnuts/adaptive_walnuts.hpp
bob-carpenter: Thanks---fixed.

include/walnuts/walnuts.hpp
bob-carpenter: That's how I started.  Then I noticed it was an exact duplicate of the code and doc with the only difference being the type `LogpGrad<F>` vs. `F`.

bob-carpenter: I'm just going to remove this counter altogether in the interest of moving on with development.

PR #18, opened 2025-07-18
Rename example using Stan, build in CI

WardBrian:
I'm not actually running it yet, because that would require building a BridgeStan model, but this should at least make sure we don't forget to update it if we change the API


WardBrian:
Well, that was instructive...

WardBrian:
@bob-carpenter besides adding the CI, the changes here will probably be necessary for you to build on clang for posteriordb testing. I found another place where GCC and clang's linker disagree with each other, which is resolved here

bob-carpenter:
Thanks.  I'm about to start doing that today.  Hopefully the rest of the code will settle down now.

examples/load_stan.hpp
```diff
@@ -143,8 +142,8 @@ class DynamicStanModel {
   }
 
   template <typename In, typename Out>
-  void constrain_draw(In &&in, Out &&out) const {
-    char *err = nullptr;
+  void constrain_draw(In&& in, Out&& out) const {
```
bob-carpenter, line 145: Thanks!  For some reason I find the `Out &&out` style very hard to read.

PR #21, opened 2025-07-28
Unit tests for combine()

WardBrian:
Adds a bunch of relatively simple tests for the combine() function on spans:

- tests that the forward/back pieces of the new span are correctly taken from the ancestor spans
- tests that the selected pieces come from the right place during acceptances and rejections
- tests some cases where barker and metropolis disagree (in particular, these would have caught https://github.com/bob-carpenter/walnuts/issues/27)


WardBrian:
Spent [some time debugging](https://github.com/google/googletest/issues/4811) and I think the Windows test failures should be resolved

bob-carpenter:
> we can't use RNG as the template type for both the base Generator (as it's called in the C++ doc) and the compound Random<RNG, S> type

"can't" -> "shouldn't"

I think it'll be confusing to use the same tparam name for both types.  I suggest we follow the doc for C++ itself and call the base uniform generator `Generator`.  Then I don't care what we call the template type for our own `Random<Generator, Scalar>` type.

My question is really whether we should just drop `Random<Generator, Scalar>` and just construct the distributions we need as we need them.  I assume they're trivial to construct from a `Generator`.  Then we can drop the `Random<Generator, Scalar>` class altogether.  Of course, that may make it harder to test because we'd need to mock a `Generator` that has the right compound value.

WardBrian:
> My question is really whether we should just drop `Random<Generator, Scalar>` and just construct the distributions we need as we need them. 

I think we shouldn't get rid of it precisely because of

> Of course, that may make it harder to test because we'd need to mock a `Generator` that has the right compound value.

(I'm also unsure they would actually be that cheap, plus we get the helper for vectors of IID normals)

bob-carpenter:
I'm not comfortable reviewing the CMake changes.  

I'll add a review for the non-Make parts.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  EXPECT_EQ(span_recieved.logp_bk_, span_expected.logp_bk_);
+
+#define EXPECT_SELECTED_POINT_EQUAL(span_recieved, input, new_logp) \
+  EXPECT_EQ(span_recieved.theta_select_, input.theta_select_);      \
+  EXPECT_EQ(span_recieved.grad_select_, input.grad_select_);        \
+  EXPECT_FLOAT_EQ(span_recieved.logp_, new_logp);
+
+// The forwards/backwards pieces of the new span are update agnostic
+template <typename T>
+class CombineSpansUpdateAgnostic : public ::testing::Test {};
```
bob-carpenter, line 80: Why is there a template parameter that's not used?

bob-carpenter: I think `::testing` can just be `testing` here---why is it qualified to start at the global namespace explicitly?  Is it guarding against some eventuality I'm failing to imagine?

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+using BothUpdates = ::testing::Types<span_test::Metropolis, span_test::Barker>;
+TYPED_TEST_SUITE(CombineSpansUpdateAgnostic, BothUpdates);
+
+TYPED_TEST(CombineSpansUpdateAgnostic, ForwardEndpointsCorrect) {
+  constexpr auto update = TypeParam::value;
```
bob-carpenter, line 86: Where is `TypeParam` coming from?  Is it a Google test constant?  I don't see it defined anywhere.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  EXPECT_EQ(span_recieved.theta_select_, input.theta_select_);      \
+  EXPECT_EQ(span_recieved.grad_select_, input.grad_select_);        \
+  EXPECT_FLOAT_EQ(span_recieved.logp_, new_logp);
+
+// The forwards/backwards pieces of the new span are update agnostic
+template <typename T>
+class CombineSpansUpdateAgnostic : public ::testing::Test {};
+
+using BothUpdates = ::testing::Types<span_test::Metropolis, span_test::Barker>;
+TYPED_TEST_SUITE(CombineSpansUpdateAgnostic, BothUpdates);
```
bob-carpenter, line 83: What's the advantage of having this stuff organized in a test suite.  I couldn't figure it out from the Google test doc for `TYPED_TEST_SUITE` in 

https://google.github.io/googletest/reference/testing.html

tests/combine_span_test.cpp
bob-carpenter: How is `TYPED_TEST` different from a regular test?  I can't understand what's going on here.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  nuts::SpanW<double> combined =
+      nuts::combine<update, nuts::Direction::Forward>(rng, std::move(span_old),
+                                                      std::move(span_new));
+
+  EXPECT_BACKWARD_ENDPOINT_EQUAL(combined, span_from);
+  EXPECT_FORWARD_ENDPOINT_EQUAL(combined, span_to);
+}
+
+TYPED_TEST(CombineSpansUpdateAgnostic, BackwardsEndpointsCorrect) {
+  constexpr auto update = TypeParam::value;
```
bob-carpenter, line 103: This variable is never used unless it's some kind of indirection through `TYPED_TEST`.  I really find macros opaque in C++.  And I can't figure out what `TypeParam` is and thus can't figure out what `value` is.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+// following tests are symmetric (same whether going forwards or backwards)
+template <typename T>
+class CombineSpansSymmetric : public ::testing::Test {};
+
+using BothDirections =
+    ::testing::Types<span_test::Forwards, span_test::Backwards>;
+TYPED_TEST_SUITE(CombineSpansSymmetric, BothDirections);
+
+TYPED_TEST(CombineSpansSymmetric, MetropolisAcceptsBarkerRejects) {
+  // cases where the two update rules differ
```
bob-carpenter, line 128: What is this line documenting?  Which two update rules, Barker and Metropolis?

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  {
+    nuts::SpanW<double> span_old = span_from;
+    nuts::SpanW<double> span_new = span_to;
+    nuts::SpanW<double> combined =
+        nuts::combine<nuts::Update::Metropolis, direction>(
+            rng, std::move(span_old), std::move(span_new));
+
+    EXPECT_SELECTED_POINT_EQUAL(combined, span_to, new_logp);
+  }
+
```
bob-carpenter, line 147: Is `lsexp` supposed to be `log_sum_exp`?  If so, spell it out everywhere.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+    nuts::SpanW<double> combined =
+        nuts::combine<nuts::Update::Barker, direction>(rng, std::move(span_old),
+                                                       std::move(span_new));
+
+    EXPECT_SELECTED_POINT_EQUAL(combined, span_from, new_logp);
+  }
+}
+
+TYPED_TEST(CombineSpansSymmetric, BarkerAndMetropolisBothAccept) {
+  // cases where the two update rules differ
```
bob-carpenter, line 197: The update rules differ but they both accept?  I find this doc confusing given the name of the test.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+    nuts::SpanW<double> span_new = span_to;
+    nuts::SpanW<double> combined =
+        nuts::combine<nuts::Update::Barker, direction>(rng, std::move(span_old),
+                                                       std::move(span_new));
+
+    EXPECT_SELECTED_POINT_EQUAL(combined, span_to, new_logp);
+  }
+}
+
+// The following test uses the same logic for all combinations of
```
bob-carpenter, line 228: I don't understand the scope of this comment---it seems to be commenting a test suite name?  Do you mean all the tests in the suite?  And what does "same logic" mean?  Same for what?  I'm OK with less doc, but I find the doc here really confusing.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+// direction and update type
+template <typename T>
+class CombineSpansUniversal : public ::testing::Test {};
+
+using AllCombinations =
+    ::testing::Types<span_test::ForwardsMetropolis,
+                     span_test::BackwardsMetropolis, span_test::ForwardsBarker,
+                     span_test::BackwardsBarker>;
+TYPED_TEST_SUITE(CombineSpansUniversal, AllCombinations);
```
bob-carpenter, line 237: This also looks very complicated.  Can we move some of this logic out of the Google test suite, which presumably more people will have to dive into the doc to understand than just using C++.  I personally find all these macros and what they're doing implicitly in scope very confusing.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+TYPED_TEST(CombineSpansUniversal, SelectedPositionAccepted) {
+  // test that the 'selected' position and gradient are as expected
```
bob-carpenter, line 240: Why is "selected" quoted here?  

Can we just name the tests something more meaningful and drop the doc?

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  constexpr auto direction = std::tuple_element_t<0, TypeParam>::value;
+  constexpr auto update = std::tuple_element_t<1, TypeParam>::value;
+
+  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
+  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);
+  double new_logp = nuts::log_sum_exp(span_from.logp_, span_to.logp_);
+
+  span_test::mock_rng rng_accept{0};  // always accept
```
bob-carpenter, line 248: This will always accept as long as we stick to finite numbers.

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+  nuts::SpanW<double> span_old = span_from;
+  nuts::SpanW<double> span_new = span_to;
+  nuts::SpanW<double> combined = nuts::combine<update, direction>(
+      rng_reject, std::move(span_old), std::move(span_new));
+
+  EXPECT_SELECTED_POINT_EQUAL(combined, span_from, new_logp);
+}
+
+// macro clean up
```
bob-carpenter, line 277: Yet another reason to dislike all these macros!

bob-carpenter (changes requested):
I did not review the CMake, which I cannot puzzle out in the amount of time I'm willing to spend on it.

As you can see, I found this PR incredibly confusing with all the macro use. I spent a couple hours on this diving into definitions of things in C++ and in Google test and feel that less reliance on the Google test macros would make the code easier to understand conditioned on the assumption that not everyone reading this is going to be a Google Test expert (I'm certainly not and still couldn't figure out parts of this and just timed myself out).  

I would like to wait to merge this until I get back to the office and can sit down with you and understand all the design decisions and alternatives here.

tests/span_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  return span;
+}
+
+/**  Mocked RNG that returns a set value deterministically */
+struct mock_rng {
+  double value;
+  double uniform_real_01() { return value; }
+};
+
```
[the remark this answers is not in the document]
WardBrian: Yes

tests/span_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+  nuts::SpanW<double> span(std::move(theta), std::move(rho), std::move(grad),
+                           logp);
+
+  span.logp_bk_ = -logp;
+
+  span.theta_bk_ = Eigen::VectorXd::Random(2);
+  span.rho_bk_ = Eigen::VectorXd::Random(2);
+  span.grad_theta_bk_ = Eigen::VectorXd::Random(2);
+
+  span.logp_fw_ = logp + 10;
```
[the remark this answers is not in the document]
WardBrian: It doesn't really matter what we do, as long as the values in each dummy span end up being unique from each other (and, ideally, from the other span in the same test) so that we can make assertions about them. If they were all the same value, comparing them for equality wouldn't be testing very much

tests/span_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+}  // namespace span_test
+
+// macros to compare parts of Spans
+#define EXPECT_FORWARD_MATCHES(span_recieved, span_expected)             \
+  EXPECT_EQ(span_recieved.theta_fw_, span_expected.theta_fw_);           \
+  EXPECT_EQ(span_recieved.rho_fw_, span_expected.rho_fw_);               \
+  EXPECT_EQ(span_recieved.grad_theta_fw_, span_expected.grad_theta_fw_); \
+  EXPECT_EQ(span_recieved.logp_fw_, span_expected.logp_fw_);
+
```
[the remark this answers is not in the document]
WardBrian: Googletest (and every other testing framework I've encountered in C++) rely on the `__FILE__` and `__LINE__` built-in macros to report errors. If you call one of their `EXPECT_...` macros inside a helper function, and it fails, then the line reported for the failure is _the helper function's body_ rather than the call site. This makes it almost useless in terms of determining which assertion failed, because they'd all point to the same lines regardless of which test they ran in.

Declaring your own macro is the only way around this, at least until there is a good enough C++20-based system that uses [std::source_location](https://en.cppreference.com/w/cpp/utility/source_location.html)

tests/combine_span_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  return span;
+}
+
+/**  Mocked RNG that returns a set value deterministically */
+struct mock_rng {
+  double value;
+  double uniform_real_01() { return value; }
+};
+
+// helpers to use GoogleTest's parameterization system
```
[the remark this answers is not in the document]
WardBrian: I think a lot of Googletest's design predates a lot of modern C++ patterns that are more common now, like heavy use of non-type template parameters (not to mention the existence of `if constexpr`).

The parameterization system is designed to parameterize over types, not generically over compile-time constants, so you need to wrap said compile time constants into types to get access to this machinery

tests/CMakeLists.txt
```diff
@@ -20,13 +24,11 @@ function(add_nuts_test TEST_NAME TEST_SOURCE)
...
-    COMMAND $<TARGET_FILE:${TEST_NAME}>
-  )
+  #  register it with CTest
+  gtest_discover_tests(${TEST_NAME})
 endfunction()
 
 add_nuts_test(online_moments_test online_moments_test.cpp)
 add_nuts_test(dual_average_test dual_average_test.cpp)
 add_nuts_test(util_test util_test.cpp)
+add_nuts_test(span_test span_test.cpp)
```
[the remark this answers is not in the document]
WardBrian: > I would've guessed the above was to find tests

The indirection here was added in https://github.com/bob-carpenter/walnuts/pull/39. All I've done here is register the new test file, and swap [`add_test`](https://cmake.org/cmake/help/latest/command/add_test.html) for [`gtest_discover_tests`](https://cmake.org/cmake/help/latest/module/GoogleTest.html#command:gtest_discover_tests). The former adds each file as a test target, while the later is googletest aware and reports each test in the file independently. 

> What does this look like for Stan

In his branch that adds cmake to Stan-Math Steve used some tricks to avoid doing this, but those have their own drawbacks.

tests/combine_span_test.cpp
WardBrian: To give some more idea of the alternatives, you have:

1. Googletest's TYPED_TEST (39 lines as formatted)

```C++
template <typename T>
class CombineSpansUpdateAgnostic : public ::testing::Test {};

using BothUpdates = ::testing::Types<span_test::Metropolis, span_test::Barker>;
TYPED_TEST_SUITE(CombineSpansUpdateAgnostic, BothUpdates);

TYPED_TEST(CombineSpansUpdateAgnostic, EndpointsCorrectWhenGoingForward) {
  constexpr auto update = TypeParam::value;

  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<update, nuts::Direction::Forward>(rng, std::move(span_old),
                                                      std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_from, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_to, combined);
}

TYPED_TEST(CombineSpansUpdateAgnostic, EndpointsCorrectWhenGoingBackward) {
  constexpr auto update = TypeParam::value;

  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<update, nuts::Direction::Backward>(rng, std::move(span_old),
                                                       std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_to, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_from, combined);
}
```

2. Copy-pasting the test body to cover all cases (57 lines)

```c++
TEST(CombineSpans, EndpointsCorrectWhenGoingForwardWithMetropolis) {
  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<nuts::Update::Metropolis, nuts::Direction::Forward>(
          rng, std::move(span_old), std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_from, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_to, combined);
}
TEST(CombineSpans, EndpointsCorrectWhenGoingForwardWithBarker) {
  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<nuts::Update::Barker, nuts::Direction::Forward>(
          rng, std::move(span_old), std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_from, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_to, combined);
}

TEST(CombineSpans, EndpointsCorrectWhenGoingBackwardWithMetropolis) {
  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<nuts::Update::Metropolis, nuts::Direction::Backward>(
          rng, std::move(span_old), std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_to, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_from, combined);
}
TEST(CombineSpans, EndpointsCorrectWhenGoingBackwardWithBarker) {
  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<nuts::Update::Barker, nuts::Direction::Backward>(
          rng, std::move(span_old), std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_to, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_from, combined);
}
```

4. Writing the test body as a template (45)

```c++
template <nuts::Update update>
void test_endpoints_correct_going_forward() {
  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<update, nuts::Direction::Forward>(rng, std::move(span_old),
                                                      std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_from, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_to, combined);
}

TEST(CombineSpans, EndpointsCorrectWhenGoingForwardWithMetropolis) {
  test_endpoints_correct_going_forward<nuts::Update::Metropolis>();
}
TEST(CombineSpans, EndpointsCorrectWhenGoingForwardWithBarker) {
  test_endpoints_correct_going_forward<nuts::Update::Barker>();
}

template <nuts::Update update>
void test_endpoints_correct_going_backward() {
  span_test::MockRandom rng{0};
  nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
  nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

  nuts::SpanW<double> span_old = span_from;
  nuts::SpanW<double> span_new = span_to;
  nuts::SpanW<double> combined =
      nuts::combine<update, nuts::Direction::Backward>(rng, std::move(span_old),
                                                       std::move(span_new));

  EXPECT_BACKWARD_ENDPOINT_EQUAL(span_to, combined);
  EXPECT_FORWARD_ENDPOINT_EQUAL(span_from, combined);
}

TEST(CombineSpans, EndpointsCorrectWhenGoingBackwardWithMetropolis) {
  test_endpoints_correct_going_backward<nuts::Update::Metropolis>();
}
TEST(CombineSpans, EndpointsCorrectWhenGoingBackwardWithBarker) {
  test_endpoints_correct_going_backward<nuts::Update::Barker>();
}
```

----

Note that this is just for a test that doesn't care about the difference between metropolis and barker, for the tests which allow any combination of `Update` and `Direction` the difference would be even more exaggerated 

I think cutting and pasting is immediately out, because if nothing else it is easy to lose track of the fact that the metropolis/barker tests are testing the same thing fundamentally. Between 1 and 3 I don't have too strong a preference

tests/span_test.cpp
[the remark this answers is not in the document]
WardBrian: It looks like Boost.UT avoids using macros (by requiring c++20): https://github.com/boost-ext/ut

tests/combine_span_test.cpp
WardBrian: Here's how it looks in Boost.UT:

```C++
suite<"combine_spans"> tests = [] {
  "endpoints_correct_going_forward"_test = []<typename U> {
    constexpr nuts::Update update = U::value;
    span_test::MockRandom rng{0.5};

    nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
    nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

    nuts::SpanW<double> span_old = span_from;
    nuts::SpanW<double> span_new = span_to;

    nuts::SpanW<double> combined =
        nuts::combine<update, nuts::Direction::Forward>(
            rng, std::move(span_old), std::move(span_new));

    expect_forward_endpoint_equal(span_to, combined);
    expect_backward_endpoint_equal(span_from, combined);
  } | std::tuple<span_test::Metropolis, span_test::Barker>();

  "endpoints_correct_going_backward"_test = []<typename U> {
    constexpr nuts::Update update = U::value;
    span_test::MockRandom rng{0.5};

    nuts::SpanW<double> span_from = span_test::dummy_span(0.3);
    nuts::SpanW<double> span_to = span_test::dummy_span(0.1);

    nuts::SpanW<double> span_old = span_from;
    nuts::SpanW<double> span_new = span_to;

    nuts::SpanW<double> combined =
        nuts::combine<update, nuts::Direction::Backward>(
            rng, std::move(span_old), std::move(span_new));

    expect_backward_endpoint_equal(span_to, combined);
    expect_forward_endpoint_equal(span_from, combined);
  } | std::tuple<span_test::Metropolis, span_test::Barker>();
};
```

note that expect_forward_endpoint_equal is a function here, not a macro, because of C++20 features.

PR #37, opened 2025-09-25
added min micro steps

bob-carpenter:
I added a minimum number of micro steps parameter.  I'm not trying to auto-tune it yet. 

There's some kind of bug in my evaluation code, I think.  But I haven't been able to track it down.  Clues:

* `WalnutsSampler` doesn't work unless `min_micro_steps = 1`
* `AdaptiveWalnutsSampler` works with just about any `min_micro_steps`.  

If `WalnutSampler` doesn't work, then `AdaptiveWalnutsSampler` shouldn't work either, as it calls `WalnutsSampler` to do the sampling.  I added prints at one point to make sure that `AdaptiveWalnutsSampler` passes the min number of micro steps and that it's being used.  So I can only imagine that I'm calling `WalnutsSampler` correctly when created in `AdaptiveWalnutsSampler` but incorrectly when called directly.  Or maybe it's just because it doesn't adapt the mass matrix, so it really struggles? 

I would like to get the `WalnutsSampler` working on its own before merging, but I'm not clear on where to go next in debugging.

### DIFF
--- a/examples/examples.cpp
+++ b/examples/examples.cpp
@@ -180,26 +180,25 @@ static void run_adaptive_walnuts(const F& target_logp_grad,
 
 int main() {
   unsigned int seed = 83435638;
-  Integer D = 200;
+  Integer D = 100;
   Integer N = 1000;
-  S step_size = 0.4;
-  Integer max_depth = 10;
-  Integer min_micro_steps = 4;
+  S step_size = 0.5;
+  Integer max_depth = 8;
+  Integer min_micro_steps = 3;
   S max_error = 1;  // 61% Metropolis
   VectorS inv_mass = VectorS::Ones(D);
   std::mt19937 rng(seed);
 
-  std::normal_distribution<S> std_normal(0, 1);
   VectorS theta_init(D);
+  std::normal_distribution<S> std_normal(0, 1);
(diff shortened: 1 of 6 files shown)


bob-carpenter:
It takes a couple draws then essentially hangs, leading to posterior standard deviations of 1e-14.  So it feels like the last bug.  It samples fine when called by `AdaptiveWalnuts`, so I don't think `Walnuts` itself is broken, just how it's getting called somewhere.  It also works when the min number of micro steps is set to 1, as it is in the tests.  Maybe it won't work without a better mass matrix?

WardBrian:
Is this for any given posterior or are you trying that correlated normal example? Nothing sticks out to me in the diff in terms of things passed in the wrong order or similar.

bob-carpenter:
Thanks for taking a look.   Just the ill-conditioned (not correlated) normal.  That one's easy to perfectly precondition, which is why adaptive walnuts is so fast there.  Maybe it's just not a great sampler with that parameter greater than 1 for this distribution.  I've stared at the chain of calls for over an hour and printed at the use destinations to make sure the right values were getting through.  So I'm tempted just to merge this and use a different example config.

PR #38, opened 2025-11-04
rhat monitor

bob-carpenter:
This is a first draft of a tool to run sampling in multiple threads and asynchronously monitor R-hat for completion.

I haven't added any doc yet in the hope that it actually makes the code a bit easier to understand.

I've more or less hit a fixed point with LLM code review.  ChatGPT 5 is still suggesting the following two relevant changes, which look reasonable.  It's still on me to slow down the controller busy spin---it has suggested like 10 different alternatives and it won't take "I'd rather spin the controller than have the chains run too long" for an answer.  It's also very offended I just let the R-hat be NaN initially and wants me to put all sorts of guards in place.  

<blockquote>
Below are targeted changes that improve allocation behavior, responsiveness, and clarity while keeping your non-blocking controller design.

### 1) Eliminate per-iteration allocations (hot path)

Reuse a per-chain scratch buffer and pass by `std::span`. This removes an `O(D)` allocation and a move per draw.

```cpp
// Sample: accept span to avoid constructing a fresh std::vector each iter
void append_draw(std::size_t iteration, double logp, std::span<const double> draw) {
  logp_.emplace_back(logp);
  theta_.insert(theta_.end(), draw.begin(), draw.end());
}

// ChainTask: reuse one buffer
class ChainTask {
  // ...
  std::vector<double> scratch_;
public:
  ChainTask(std::size_t chain_id, std::size_t draws_per_chain, Sampler& sampler,
            Queue& q, std::latch& start_gate)
      : chain_id_(chain_id), draws_per_chain_(draws_per_chain), sampler_(sampler),
        sample_(chain_id, sampler.dim(), draws_per_chain),
        q_(q), start_gate_(start_gate), scratch_(sampler.dim()) {}

  void operator()(std::stop_token st) {
    initiated_qos();
    start_gate_.arrive_and_wait();
    for (std::size_t iter = 0; iter < draws_per_chain_; ++iter) {
      double logp = sampler_(scratch_);                    // changed
      logp_stats_.push(logp);
      sample_.append_draw(iter, logp, scratch_);           // changed
      q_.emplace(logp_stats_.count(), logp_stats_.mean(),  // unchanged
                 logp_stats_.sample_variance());
      if (st.stop_requested()) break;
    }
  }
  // ...
};
```

```cpp
// StandardNormalSampler: fill into caller-provided buffer
double operator()(std::span<double> out) noexcept {
  double log_density = 0.0;
  for (std::size_t i = 0; i < dim_; ++i) {
    double x = normal_dist_(engine_);
    out[i] = x;
    log_density += -0.5 * x * x;
  }
  return log_density;
}
```

### 2) Make the ring buffer explicitly SPSC and cheaper to index

Use a power-of-two capacity and a bit mask; store elements in `std::array` to avoid default-constructing/relocating a `std::vector<T>` payload.

```cpp
template <class T, std::size_t CapacityPow2>
class alignas(std::hardware_destructive_interference_size) SpscRingBuffer {
  static_assert(std::has_single_bit(CapacityPow2),
                "Capacity must be a power of two.");
public:
  bool emplace(const T& x) noexcept {
    auto w = write_.load(std::memory_order_relaxed);
    auto next = (w + 1) & (CapacityPow2 - 1);
    if (next == read_.load(std::memory_order_acquire)) return false;
    data_[w] = x;                                         // trivially cheap
    write_.store(next, std::memory_order_release);
    return true;
  }
  template <class... Args>
  bool emplace(Args&&... args) noexcept {
    auto w = write_.load(std::memory_order_relaxed);
    auto next = (w + 1) & (CapacityPow2 - 1);
    if (next == read_.load(std::memory_order_acquire)) return false;
    data_[w] = T(std::forward<Args>(args)...);
    write_.store(next, std::memory_order_release);
    return true;
  }
  bool pop(T& out) noexcept {
    auto r = read_.load(std::memory_order_relaxed);
    if (r == write_.load(std::memory_order_acquire)) return false;
    out = std::move(data_[r]);
    read_.store((r + 1) & (CapacityPow2 - 1), std::memory_order_release);
    return true;
  }
  constexpr std::size_t capacity() const noexcept { return CapacityPow2; }
private:
  std::array<T, CapacityPow2> data_{};
  alignas(std::hardware_destructive_interference_size) std::atomic<std::size_t> read_{0};
  alignas(std::hardware_destructive_interference_size) std::atomic<std::size_t> write_{0};
};

// Replace aliases
constexpr std::size_t RING_CAPACITY = 64;  // generous to avoid producer back-pressure
using Queue = SpscRingBuffer<SampleStats, RING_CAPACITY>;
```
</blockquote>


examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+  void operator()(std::stop_token st) {
+    initiated_qos();
+    start_gate_.arrive_and_wait();
+    for (std::size_t iter = 0; iter < draws_per_chain_; ++iter) {
+      auto [logp, theta] = sampler_();
+      logp_stats_.push(logp);
+      sample_.append_draw(logp, std::move(theta));
+      q_.emplace(logp_stats_.count(), logp_stats_.mean(),
+                 logp_stats_.sample_variance());
```
WardBrian: It feels like this could be simpler if you had a function on the welford accumulator that returned a SampleStats

WardBrian (commented):
Overall I find this pretty easy to read, but a few things did jump out at me

WardBrian:
> I can't figure out why the behavior is different if I explicitly join. My understanding is that jthreads are implicitly joined at the end of scope. I need to understand this much better.

I figured it out -- unless you put a join there, the scope ends basically immediately after the threads are created, and that triggers the stop token being set in the controller thread. So your options are:
1. call controller.join()
2. just ignore the stop token in the controller thread, make it `while (true)`. You don't really want that thread to be cancellable, so I prefer this option honestly.

You can then delete that for loop

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+private:
+  std::size_t chain_id_;
+  std::size_t D_;
+  std::size_t Nmax_;
+  std::vector<double> theta_;
+  std::vector<double> logp_;
+};
+
+// see https://rigtorp.se/ringbuffer/
```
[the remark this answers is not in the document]
bob-carpenter: Will add @SteveBronder's version later or maybe this one.

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,387 @@
...
+  {  // scope for jthread RAII control
+    std::vector<std::jthread> workers;
+    workers.reserve(M);
+    for (std::size_t m = 0; m < M; ++m) {
+      workers.emplace_back(std::ref(tasks[m]));
+    }
+    std::jthread controller([&](std::stop_token st) {
+      interactive_qos();
+      controller_loop(st, queues, workers, rhat_threshold, start_gate);
+    });
```
WardBrian: is there a reason to run the controller thread on its own thread rather than on the main thread here?

examples/rhat_monitor.cpp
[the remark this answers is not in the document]
SteveBronder: Rigtorp's SPSCQueue is very nice. We can use his simpler one just copy/pasted if we do not want another dependency.

One reason we may want our own is so that we can change it to overwrite old values that have not been read from the master thread yet.

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+  std::size_t capacity() const noexcept { return Capacity; }
+
+ private:
+  std::vector<T> data_;
+  alignas(std::hardware_destructive_interference_size)
+      std::atomic<std::size_t> read_idx_{0};
+  alignas(std::hardware_destructive_interference_size)
+      std::atomic<std::size_t> write_idx_{0};
+};
```
SteveBronder: One thing that would be nice here is a `bulk_pop` that would flush out the whole queue. Flushing out the queue would be nice since we can get a bunch of stats at once from a chain. We could essentially have a check where we ask if the queue is full and if so it does the bulk pop. I think it would work something like the following
1. Ask queue if it is full
2. If queue is full, call `bulk_pop passing in an array of size `Capacity`
3. Set `write_idx_` to `read_idx_  - 1` to lock the ring buffer
4. swap the array we passed in with the array in `data_`
5. reset `write_idx_` and `read_idx_` to zero to unlock the ring buffer

The time we lock out writes should be very short with this impl since we are just doing a swap. And since we would only do it if the queue is full anyway the writes would have been rejected anway.

SteveBronder (commented):
Few comments. Nothing major! If there is anything you want me to look at that I missed lmk happy to review again

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+      : dim_(dim), engine_(seed), normal_dist_(0, 1) {}
+
+  void sample(std::vector<double>& draw, double& lp) noexcept {
+    lp = 0;
+    for (std::size_t i = 0; i < dim_; ++i) {
+      double x = normal_dist_(engine_);
+      draw.push_back(x);
+      lp += -0.5 * x * x;  // unnomalized
+    }
+  }
```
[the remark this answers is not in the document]
bob-carpenter: I changed it to return the log density---it doesn't need an init.  

Note:  I only did it the way I did it initially to parallel Stan's autodiff functionals.

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+ public:
+  WelfordAccumulator() : n_(0), mean_(0.0), M2_(0.0) {}
+
+  void push(double x) {
+    ++n_;
+    const double delta = x - mean_;
+    mean_ += delta / static_cast<double>(n_);
+    const double delta2 = x - mean_;
+    M2_ += delta * delta2;
+  }
```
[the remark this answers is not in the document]
bob-carpenter: I used `observe`.  That matches what I do in the rest of the adaptation code.

examples/rhat_monitor.cpp
bob-carpenter: I said this below in the top-level comment, but I'll paraphrase here.

What's getting pushed into the ring buffers by the chains is the latest mean and variance estimate for the whole chain.  That means we only ever want to read the last thing that pushed.  So we don't need to bulk pop as we only ever need the latest.

We should talk about how to support this pattern most efficiently, but I'm going to mark this conversation as resolved.

examples/rhat_monitor.cpp
```diff
@@ -270,26 +235,24 @@ void debug_print(double variance_of_means, double mean_of_variances,
...
   std::vector<std::size_t> counts(M, 0);
   while (true) {
     for (std::size_t m = 0; m < M; ++m) {
-      ChainStats u;
-      while (queues[m].pop(u)) {
-        chain_means[m] = u.sample_mean;
-        chain_variances[m] = u.sample_var;
-        counts[m] = u.count;
-      }
+      ChainStats u = chain_statses[m].load();
```
SteveBronder: Without some flag here to tell the main thread it has taken the workers info already you could potentially be pulling in old info that you have already taken

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,367 @@
...
+#include <iostream>
+#include <latch>
+#include <numeric>
+#include <random>
+#include <stop_token>
+#include <string>
+#include <thread>
+#include <vector>
+
+#define VERY_INLINE [[gnu::always_inline]] inline
```
SteveBronder: It is annoying to write, but generally its good to check if the attribute exists before making the macro.
```suggestion
#if defined __has_attribute
#if __has_attribute(always_inline)
#ifndef WALNUTS_STRONG_INLINE
#define WALNUTS_STRONG_INLINE [[gnu::always_inline]] inline
#endif
#endif
#endif
```

bob-carpenter (commented):
Because this is orthogonal to the rest of the code base and just adds four new files that are not used elsewhere, I'm just going to merge to cut down on the number of branches floating around.  We'll probably strip out some of the examples cruft before releasing.

examples/adapt_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+  }
+}
+
+
+template <typename Sampler>
+AdaptResult sample(const AdaptConfig& cfg, std::vector<Sampler>& samplers) {
+  std::vector<PaddedBuffer> buffers = construct_buffers(cfg.num_chains, cfg.dim);
+
+  std::latch start_gate(static_cast<std::ptrdiff_t>(cfg.num_chains));
+  std::stop_source stop_source;
```
WardBrian, line 287: unused variable

examples/adapt_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+    for (auto& r : runners) {
+      r->request_stop();
+    }
+  };
+  
+  AdaptResult res = controller_loop(buffers, cfg, stop_all);
+
+  for (auto& r : runners) {
+    r->join();
+  }
```
WardBrian, line 307: Because these are ultimately jthreads, you shouldn't need to manually join them

examples/adapt_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+
+template <typename Sampler>
+AdaptResult sample(const AdaptConfig& cfg, std::vector<Sampler>& samplers) {
+  std::vector<PaddedBuffer> buffers = construct_buffers(cfg.num_chains, cfg.dim);
+
+  std::latch start_gate(static_cast<std::ptrdiff_t>(cfg.num_chains));
+  std::stop_source stop_source;
+
+  std::vector<std::unique_ptr<AdaptRunner<Sampler>>> runners;
```
WardBrian, line 289: I don't understand why these are behind a unique_ptr?

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+  auto stop_all = [&] {
+    for (auto& r : runners) {
+      r.request_stop();
+    }
+  };
+  controller_loop(stats_by_chain, rhat_threshold, start_gate,
+                  max_draws_per_chain, stop_all);
+  for (auto& r : runners) {
+    r.join();  // avoid race before taking records
+  }
```
WardBrian, line 358: I have the same feeling about manually joining jthreads reading as odd, but here it does make sense. You could wrap the preceeding code in a local block to get the same effect without this.

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+  std::ofstream out(path, std::ios::binary);
+  if (!out) {
+    throw std::runtime_error("could not open file: " + path);
+  }
+  std::vector<char> filebuf(8u << 20); //  8 MB buffer bigger than usual
+  out.rdbuf()->pubsetbuf(filebuf.data(), static_cast<std::streamsize>(filebuf.size()));
+  write_fixed_header(out, dim, chains);
+  write_column_names(out, dim);
+  write_chains(out, dim, chains);
```
WardBrian, line 435: I assume this binary output is temporary, but I wanted to note that it's probably a good idea to pad the header such that the draws start at an 8 byte boundary, as this makes things like mmapping the file easier/faster.

I did something like this when I was hacking binary output on top of cmdstan, see https://github.com/WardBrian/cmdstan/blob/basic-binary-output/src/cmdstan/binary_writer.hpp#L102-L114

examples/adapt_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+  std::uint32_t chain_id_;
+  AdaptConfig cfg_;
+  Buffer& buffer_;
+  std::latch& start_gate_;
+  AdaptiveSampler& adapter_;
+};
+
+template <class AdaptiveSampler>
+class AdaptRunner {
```
WardBrian, line 127: This class seems a bit weird to me -- I'd probably not give it a name and just use `jthread<AdaptWorker>` where appropriate, unless it's serving some extra purpose I'm missing

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+  std::size_t draws_per_chain_;
+  Sampler& sampler_;
+  ChainRecord chain_record_;
+  WelfordAccumulator logp_stats_;
+  AtomicChainStats& acs_;
+  std::latch& start_gate_;
+};
+
+template <class Sampler>
+class ChainRunner {
```
WardBrian, line 246: Similar to my above comment, I'd be curious about the benefit of this class as opposed to directly using `jthread<ChainWorker>`

WardBrian (commented):
A few things stood out to me, but I think overall this code in the examples folder looks pretty reasonable. I'd recommend the next step be separating into various files, e.g. all the `elementwise_` functions over spans going into a `walnuts/math.hpp` etc. 

Once I had skimmed those there isn't really that much code left in the end, which is nice.

I'll leave Steve to comment on the padding/atomics/buffer code. Nothing immediately jumped out at me there, but I'm not all that familiar

examples/adapt_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+#include <span>
+#include <stop_token>
+#include <thread>
+#include <vector>
+
+#include "walnuts/padded.hpp"
+#include "walnuts/triple_buffer.hpp"
+
+
+struct AdaptSnapshot {
```
SteveBronder, line 25: I would force 64 byte alignment here to prevent false sharing due to cache alignment.
```suggestion
struct alignas(64) AdaptSnapshot {
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+#include <thread>
+#include <vector>
+
+#include "walnuts/padded.hpp"
+
+#if defined(__clang__) || defined(__GNUC__)
+#define VERY_INLINE __attribute__((always_inline)) inline
+#else
+#define VERY_INLINE inline
+#endif
```
SteveBronder, line 32: A little extra, but I recommend using `__has_attribute` and putting `WALNUTS` in front of your macros
```suggestion
#ifdef __has_attribute
#if __has_attribute(always_inline)
#define WALNUTS_ALWAYS_INLINE_ __attribute__((always_inline)) inline
#else
#define WALNUTS_ALWAYS_INLINE_ inline
#endif
#endif
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+  }
+
+  ChainRecord take_record() && { return std::move(chain_record_); }
+
+ private:
+  std::size_t draws_per_chain_;
+  Sampler& sampler_;
+  ChainRecord chain_record_;
+  WelfordAccumulator logp_stats_;
+  AtomicChainStats& acs_;
```
SteveBronder, line 241: I don't like raw references as it can make copy/move constructors accidentally break due to the reference falling out of scope. Use `std::reference_wrapper` and then have access methods


```suggestion
  std::reference_wrapper<AtomicChainStats> acs_;
```

(in public)
```cpp
ALWAYS_INLINE_ auto& acs() {
  return acs_.get();
}
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+  ChainWorker(std::size_t draws_per_chain, Sampler& sampler,
+              AtomicChainStats& acs, std::latch& start_gate)
+      : draws_per_chain_(draws_per_chain),
+        sampler_(sampler),
+        chain_record_(sampler.dim(), draws_per_chain),
+        acs_(acs),
+        start_gate_(start_gate) {}
+
+  void operator()(std::stop_token st) {
+    static constexpr std::size_t YIELD_PERIOD = 64;
```
SteveBronder, line 217: I would have this be settable by the user for benchmarking. I think this will be pretty dependent on the workload

If you do want it to be a compile time constant then I would make a `config.hpp` and have a `ChainWorker` namespace that holds compile time attributes. 

```cpp
namespace Config {
namespace ChainWorker {
  static constexpr std::size_t YIELD_PERIOD = 64;
}
}
```
Then you access it via `config::ChainWorker::YIELD_PERIOD`

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+ public:
+  ChainRecord(std::size_t D, std::size_t Nmax)
+      : D_(D), Nmax_(Nmax), n_(0), theta_(Nmax * D), logp_(Nmax) {}
+
+  std::span<double> draw(std::size_t n) noexcept {
+    return {&theta_[n * D_], D_};
+  }
+
+  std::span<const double> draw(std::size_t n) const noexcept {
+    return {&theta_[n * D_], D_};
```
SteveBronder, line 112: (optional)
```suggestion
    return {theta_.data() + n * D_, D_};
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+    data_.store(init, std::memory_order_relaxed);
+  }
+
+  void store(const ChainStats& p) noexcept {
+    data_.store(p, std::memory_order_relaxed);  //  // release);
+  }
+
+  ChainStats load() const noexcept {
+    return data_.load(std::memory_order_relaxed);  // acquire);
+  }
```
SteveBronder, line 96: I think it would be nice to let users choose here

```suggestion
  void store(const ChainStats& p, std::memory_order mem_order = std::memory_order::relaxed) noexcept {
    data_.store(p, mem_order);  //  // release);
  }

  ChainStats load(std::memory_order mem_order = std::memory_order::relaxed) const noexcept {
    return data_.load(mem_order);  // acquire);
  }
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+      auto draw = chain_record_.draw(iter);
+      auto& lp = chain_record_.logp(iter);
+      lp = sampler_.sample(draw);
+      chain_record_.commit();
+      logp_stats_.observe(lp);
+      acs_.store(logp_stats_.sample_stats());
+    }
+  }
+
+  ChainRecord take_record() && { return std::move(chain_record_); }
```
SteveBronder, line 234: If you are okay using C++23 you can use ["deducing `this`"](https://devblogs.microsoft.com/cppblog/cpp23-deducing-this/) here in a nice way


```suggestion
  template <typename Self>
  ChainRecord record(this Self&& self) { return std::forward<Self>(self).chain_record_; }
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+ private:
+  std::size_t draws_per_chain_;
+  Sampler& sampler_;
```
SteveBronder, line 238: See comment about std reference wrapper [here](https://github.com/flatironinstitute/walnuts/pull/38/changes#r2677552547)

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+      : draws_per_chain_(draws_per_chain),
+        sampler_(sampler),
+        chain_record_(sampler.dim(), draws_per_chain),
+        acs_(acs),
+        start_gate_(start_gate) {}
+
+  void operator()(std::stop_token st) {
+    static constexpr std::size_t YIELD_PERIOD = 64;
+    interactive_qos();
+    start_gate_.arrive_and_wait();
```
SteveBronder, line 219: What is the reason to wait here?

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+  ChainRunner(const ChainRunner&) = delete;
+  ChainRunner& operator=(const ChainRunner&) = delete;
+  ChainRunner(ChainRunner&&) noexcept = default;
+  ChainRunner& operator=(ChainRunner&&) noexcept = default;
+
+  void request_stop() { thread_.request_stop(); }
+
+  void join() { thread_.join(); }
+
+  ChainRecord take_record() { return std::move(worker_).take_record(); }
```
SteveBronder, line 262: From my comment [here](https://github.com/flatironinstitute/walnuts/pull/38/changes#r2677593561)
```suggestion
  template <typename Self>
  auto record(this Self&& self) { return std::forward<Self>(self).worker_.record(); }
```

The rvalue'ness of `self` will trickle down to the call to `record()`

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+  std::cout << "RHAT: " << r_hat << "  NUM_DRAWS: " << num_draws
+            << "  COUNTS: ";
+  for (std::size_t m = 0; m < M; ++m) {
+    if (m > 0) {
+      std::cout << ", ";
+    }
+    std::cout << counts[m];
+  }
+  std::cout << std::endl;
+}
```
SteveBronder, line 282: Use chatgpt or claude to make a nice jsonl printer for debugging. So then it write each json line to a file and it's easier for debugging logs.

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+    ++num_rhat_evals;
+    // debug_print(variance_of_means, mean_of_variances, num_draws, r_hat, counts);
+
+    std::cout << std::setprecision(6) << std::fixed
+	      << std::setw(8) << std::setfill(' ')
+	      << '\r' // begin of currnet line
+	      << "R hat " << r_hat
+	      << "  #draws " << num_draws
+	      << std::flush;
```
SteveBronder, line 321: delete

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+    
+    if (r_hat <= rhat_threshold || num_draws == M * max_draws_per_chain) {
+      break;
+    }
+    std::this_thread::sleep_until(next);
```
SteveBronder, line 326: Why are you sleeping here?

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+    next += PERIOD;
+  }
+  std::cout << "\n# Rhat evals by controller: " << num_rhat_evals << std::endl;
```
SteveBronder, line 329:
```suggestion
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+}
+
+static inline void write_string(std::ostream& out, const std::string& s) {
+  write_u32(out, s.size());
+  out.write(s.data(), static_cast<std::streamsize>(s.size()));
+}
+
+static void write_fixed_header(std::ofstream& out, std::size_t dim,
+                               const std::vector<ChainRecord>& chains) {
+  static constexpr std::array<char, 8> NAME = {'W','A','L','N','U','T','S','\0'};
```
SteveBronder, line 386:
```suggestion
  static constexpr std::string_view NAME = "WALNUTS";
```

examples/rhat_monitor.cpp
```diff
@@ -0,0 +1,509 @@
...
+  static constexpr std::uint32_t VERSION = 1;
```
SteveBronder, line 387: put in `config.hpp`

examples/adapt_monitor.cpp
```diff
@@ -0,0 +1,404 @@
...
+#include <limits>
+#include <memory>
+#include <random>
+#include <span>
+#include <stop_token>
+#include <thread>
+#include <vector>
+
+#include "walnuts/padded.hpp"
+#include "walnuts/triple_buffer.hpp"
```
SteveBronder, line 22: (optional) I just think `<>` is prettier
```suggestion
#include <walnuts/padded.hpp>
#include <walnuts/triple_buffer.hpp>
```

include/walnuts/triple_buffer.hpp
```diff
@@ -0,0 +1,125 @@
...
+  /** Contiguous memory to store the buffer values.  */
+  std::array<T, 3> buffers_;
+
+  /** Current front of the buffer, aligned to avoid cache line conflicts. */
+  alignas(walnuts::DI_SIZE) std::atomic<index_t> front_;
+
+  /** Current spare buffer, aligned to avoid cache line conflicts. */
+  alignas(walnuts::DI_SIZE) std::atomic<index_t> spare_;
+
+  /** Current back buffer---only used by writer and publisher. */
```
SteveBronder, line 118: Is there a publisher thread that is separate from the writer thread? If not I would make it clear this is meant for SPSC

include/walnuts/triple_buffer.hpp
```diff
@@ -0,0 +1,125 @@
...
+  index_t back_;
+
+  /** Current read buffer---only used by reader. */
+  index_t read_;
```
SteveBronder, line 122: `back_` is read by the publisher and `read_` is read by the reader, so you want to make sure these are on different cache lines (like the rigtorp SPSC queue)

```suggestion
  alignas(walnuts::DI_SIZE) index_t back_;

  /** Current read buffer---only used by reader. */
  alignas(walnuts::DI_SIZE) index_t read_;
```

include/walnuts/triple_buffer.hpp
```diff
@@ -0,0 +1,125 @@
...
+   */
+  const T& read_latest() noexcept {
+    const int idx = front_.load(std::memory_order_acquire);
+    if (idx != read_) {
+      const int old = read_;
+      read_ = idx;
+      spare_.store(old, std::memory_order_release);
+    }
+    return buffers_[read_];
+  }
```
SteveBronder, line 103: This has a race condition. Below is a copy/paste of a google sheet iterating through publish running twice

<google-sheets-html-origin><style type="text/css"></style>
Step | Op | publish_ | front_ | spare_ | back_ | read_
-- | -- | -- | -- | -- | -- | --
0 | Init | NA | 0 | 1 | 2 | 0
1 | p:publish_= back_ | 2 | 0 | 1 | 2 | 0
2 | p:spare.exch(publish) | 2 | 0 | 2 | 2 | 0
3 | back_ = | 2 | 0 | 2 | 1 | 0
4 | front_.store(publish) | 2 | 2 | 2 | 1 | 0
5 | p:publish_= back_ | 1 | 2 | 2 | 1 | 0
6 | p:spare.exch(publish) | 1 | 2 | 1 | 1 | 0
7 | back_ = | 1 | 2 | 1 | 2 | 0
8 | front_.store(publish) | 1 | 1 | 1 | 2 | 0
9 | p:publish_= back_ | 2 | 1 | 1 | 2 | 0
10 | p:spare.exch(publish) | 2 | 1 | 2 | 2 | 0
11 | back_ = | 2 | **1** | 2 | **1** | 0
12 | front_.store(publish) | 2 | 2 | 2 | 1 | 0

The offending line is line 11, where `front_` and `back_` can be the same value. If the producer is calling `publish()` on this line at the same time as the reader thread is calling `read_latest()` then `read_` ticks up to `front_`, but `front_` and `back_` are also both `1`. So now the reader and writer are both writing to the same block

https://docs.google.com/spreadsheets/d/1D6J6XOGU2YqTzpleP61XGf58SCq50SkEUbLlyF_nM_I/edit?usp=sharing


Claude is saying you can just swap the order of back_ and front_ storage, but I need to work out whether that fixes this or not. 


```suggestion
  void publish() noexcept {
    const int published = back_;
    front_.store(published, std::memory_order_release);
    back_ = spare_.exchange(published, std::memory_order_acq_rel);
  }

  /**
   * Return a constant reference to the latest value.
   *
   * @return The latest value.
   */
  const T& read_latest() noexcept {
    const int idx = front_.load(std::memory_order_acquire);
    if (idx != read_) {
      const int old = read_;
      read_ = idx;
      spare_.store(old, std::memory_order_release);
    }
    return buffers_[read_];
  }
```

examples/rhat_monitor.cpp
WardBrian: I would avoid C++23 for the time being, I think it will make wrapping this code up in Python more annoying on some platforms

examples/adapt_monitor.cpp
bob-carpenter: That's only when they go out of scope, right?  I was thinking this wouldn't have them going out of scope, but I think it does, so I removed the join.

examples/adapt_monitor.cpp
bob-carpenter: There's no nullity constructor for `AdaptRunner`.  If you can figure out how to make this compile, it's fine to have the container own the memory.  I don't otherwise need to move the pointers around.

examples/rhat_monitor.cpp
bob-carpenter: They were like this because I need the records afterward, and they're owned by the embedded `ChainWorker`.  I refactored to preallocate.

examples/adapt_monitor.cpp
bob-carpenter: Good catch---I think it was residue from refactoring---much simpler as it lets me also get rid of the pointers.

examples/adapt_monitor.cpp
bob-carpenter: I just got rid of the runner as @WardBrian suggests below---much cleaner and no more pointers.

examples/rhat_monitor.cpp
bob-carpenter: I cribbed from `rhat_monitor.cpp` to build `adapt_monitor.cpp`.  I think I started with a big compound class, factored it apart, then was left with this vestigial class.  Algebraically it feels like leaving a `1 *` or `0 + ` lying around.

bob-carpenter: I was also able to cleanly preallocate without a lot of indirection and get rid of the wrapper class.

examples/rhat_monitor.cpp
bob-carpenter: Cool, and thanks for giving me the heads up about C++23, @WardBrian.  I'll skip the deduction for now and start getting impatient about my next C++ version :-)

examples/rhat_monitor.cpp
bob-carpenter: Thanks, but I'm not taking anything any more---I just preallocated and passed in for use so the ownership is higher up.

examples/rhat_monitor.cpp
bob-carpenter: It was just for debugging and I'd already commented it out, so I just removed it.

examples/rhat_monitor.cpp
bob-carpenter: Not necessary for correctness.  It's just given this is intended to be a user-facing app, one convergence check per 0.01s seems enough.  If it's taken longer than that to do a convergence check, the sleeper wakes up right away (hopefully `sleep_until` never actually sleeps).

examples/rhat_monitor.cpp
bob-carpenter: I like being able to see the R-hat count down and number of draws count up until it stops.  Eventually, we'll want to get rid of all uses of `std::cout`, etc., other than in the CLI.

I'd like to get some feedback on how to report progress and make this easy to set up to report progress.  The controller is the one fixed component---all the other progress is by chain.

examples/rhat_monitor.cpp
bob-carpenter: I added this as an out-argument to the sampler to get rid of `std::cout`.

examples/rhat_monitor.cpp
bob-carpenter: Nice---I didn't know about `string_view`!

include/walnuts/triple_buffer.hpp
bob-carpenter: No.  I fixed the comments.  I was originally confused because I had a copy of this file floating around in the original wrong directory where I put it.

include/walnuts/triple_buffer.hpp
bob-carpenter: I made the swap, but I'm going to leave this one unresolved as a placeholder.  And just to double check---there weren't any changes to `read_latest()`?

SteveBronder: I have to work this out, but my general rule of thumb is that it is unsafe to take a reference to an object inside of a concurrent object. After you change `spare_` I think publish could take `spare_` as `back_`, no? I would do a `swap_latest` here that takes swap a thread local `T` with the buffers `T` before changing `spare_`


```suggestion
  void swap_latest(T& out) noexcept {
    const int idx = front_.load(std::memory_order_acquire);
    if (idx != read_) {
      const int old = read_;
      read_ = idx;
      out.swap(buffers_[read_])
      spare_.store(old, std::memory_order_release);
    } else {
      out.swap(buffers_[read_])
    }
  }
```

SteveBronder: I think `read_latest()` is fine

PR #44, opened 2025-12-11
Min micro [WIP]

bob-carpenter:
This is adding tuning for the minimum number of micro steps.  It's not ready to be merged yet, but I wanted a sense of the diff so far as it's not big but a lot of things changed around it.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -2,6 +2,7 @@ if(MSVC)
   set(gtest_force_shared_crt on)
 endif()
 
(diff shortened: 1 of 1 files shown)


WardBrian:
I tried for about an hour to see if I could easily port the history from being branched off `benchmarks` to `main` and unfortunately I just don't see a way to do it that's better than just re-writing the code on the correct base commit. There are two many intermediate merges in the way and other oddities, like `main` and `benchmarks` both having implementations of Adam but they seem to be two different disjoint histories which means there are a lot of merge conflicts and it's not even clear which one should be cherry-picked

bob-carpenter:
Superseded by #45 that fixed the history

PR #47, opened 2026-01-05
Fix swapped max_step_halvings/min_micro_steps call

### LINKED ISSUE #46: Switched args in `build_leaf` fun calling `macro_step` in walnuts.hpp
Unless I'm missing something it appears that the order of the arguments for 4 and 5 is switched from the definition. 

https://github.com/flatironinstitute/walnuts/blob/9e8acf5bc5a5e23c34638f47b73c1549a6330df9/include/walnuts/walnuts.hpp#L337 

It should be `max_step_halvings` as arg 4 and `min_micro_steps` as per the definition in macro step https://github.com/flatironinstitute/walnuts/blob/9e8acf5bc5a5e23c34638f47b73c1549a6330df9/include/walnuts/walnuts.hpp#L216.

WardBrian:
Closes #46

### DIFF
--- a/include/walnuts/walnuts.hpp
+++ b/include/walnuts/walnuts.hpp
@@ -334,8 +334,8 @@ std::optional<SpanW<S>> build_leaf(const F& logp_grad, const SpanW<S>& span,
   Vec<S> rho_next;
   Vec<S> grad_theta_next;
   S logp_theta_next;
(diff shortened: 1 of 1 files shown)


bob-carpenter (approved):
D'oh!  This is the danger of unnamed arguments of the same type.

PR #48, opened 2026-01-12
Refactor monitors

bob-carpenter:
I believe I've addressed all the comments from @WardBrian and @SteveBronder at

https://github.com/flatironinstitute/walnuts/pull/38

and wrapped them up in this PR.  Sorry for the awkward process here.  And thanks so much for all the constructive suggestions---it really tightened everything up nicely.

### DIFF
--- a/examples/rhat_monitor.cpp
+++ b/examples/rhat_monitor.cpp
@@ -207,10 +207,11 @@ template <class Sampler>
 class ChainWorker {
  public:
   ChainWorker(std::size_t draws_per_chain, Sampler& sampler,
+	      ChainRecord& chain_record,
               AtomicChainStats& acs, std::latch& start_gate)
       : draws_per_chain_(draws_per_chain),
         sampler_(sampler),
-        chain_record_(sampler.dim(), draws_per_chain),
+        chain_record_(chain_record), // sampler.dim(), draws_per_chain),
         acs_(acs),
         start_gate_(start_gate) {}
 
@@ -232,39 +233,13 @@ class ChainWorker {
     }
   }
 
-  ChainRecord take_record() && { return std::move(chain_record_); }
-
- private:
+private:
   std::size_t draws_per_chain_;
   Sampler& sampler_;
-  ChainRecord chain_record_;
+  ChainRecord& chain_record_;
   WelfordAccumulator logp_stats_;
   AtomicChainStats& acs_;
   std::latch& start_gate_;
-};
-
-template <class Sampler>
-class ChainRunner {
- public:
-  ChainRunner(std::size_t draws_per_chain, Sampler& sampler,
-              AtomicChainStats& acs, std::latch& start_gate)
-      : worker_(draws_per_chain, sampler, acs, start_gate),
-        thread_(std::ref(worker_)) {}
-
-  ChainRunner(const ChainRunner&) = delete;
-  ChainRunner& operator=(const ChainRunner&) = delete;
-  ChainRunner(ChainRunner&&) noexcept = default;
-  ChainRunner& operator=(ChainRunner&&) noexcept = default;
-
-  void request_stop() { thread_.request_stop(); }
-
-  void join() { thread_.join(); }
-
-  ChainRecord take_record() { return std::move(worker_).take_record(); }
-
- private:
-  ChainWorker<Sampler> worker_;
-  std::jthread thread_;
 };
 
 void debug_print(double variance_of_means, double mean_of_variances,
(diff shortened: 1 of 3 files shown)


WardBrian (approved):
Thanks, the parts I looked at now all read very naturally to me!

examples/rhat_monitor.cpp
```diff
@@ -206,98 +213,62 @@ template <class Sampler>
...
         std::this_thread::yield();
       }
-      auto draw = chain_record_.draw(iter);
-      auto& lp = chain_record_.logp(iter);
-      lp = sampler_.sample(draw);
-      chain_record_.commit();
+      auto draw = chain_record_.get().draw(iter);
+      auto& lp = chain_record_.get().logp(iter);
+      lp = sampler_.get().sample(draw);
+      chain_record_.get().commit();
```
SteveBronder, line 238: (optional) I feel like the `get` here can be wordy and like having accessor member functions

examples/rhat_monitor.cpp
```diff
@@ -206,98 +213,62 @@ template <class Sampler>
...
-  std::cout << "RHAT: " << r_hat << "  NUM_DRAWS: " << num_draws
-            << "  COUNTS: ";
-  for (std::size_t m = 0; m < M; ++m) {
-    if (m > 0) {
-      std::cout << ", ";
-    }
-    std::cout << counts[m];
-  }
-  std::cout << std::endl;
-}
```
SteveBronder, line 282: (not for this PR)
I know you want to be dependency lite, but I've been using Quill lately and it's been very nice!

https://quillcpp.readthedocs.io/en/latest/quick_start.html

It has nice things that let you turn on and off logging at certain levels (critical, error, info, warning, debug) at compile time and runtime.

It even has nice things like `LOG_DEBUG_LIMIT` which will only print at only once for a user defined period.

```
// container types like `count` can be passed directly to the logger
// prints info once every 1 second
LOG_INFO_LIMIT(std::chrono::seconds{1}, logger, "RHAT: {} NUM_DRAWS: {} COUNTS: {}", r_hat, num_draws, counts);
```

Adding this would be a very separate PR

examples/rhat_monitor.cpp
```diff
@@ -323,75 +294,74 @@ static void controller_loop(std::vector<walnuts::Padded<AtomicChainStats>>& stat
...
+template <typename Out>
+static inline void write_string(Out& out, const std::string& s) {
   write_u32(out, s.size());
   out.write(s.data(), static_cast<std::streamsize>(s.size()));
 }
 
-static void write_fixed_header(std::ofstream& out, std::size_t dim,
+template <typename Out>
+static void write_fixed_header(Out& out, std::size_t dim,
                                const std::vector<ChainRecord>& chains) {
```
SteveBronder, line 354: (optional) could do some nice concepts things here

```cpp
/**
 * In `concepts.hpp`
 */
/// @brief Concept for 32-bit signed integers
template <typename T>
concept SignedInteger32 =
    sizeof(T) == 4 && std::is_signed_v<T> && std::is_integral_v<T>;

/// @brief Concept for 32-bit unsigned integers
template <typename T>
concept UnsignedInteger32 =
    sizeof(T) == 4 && !std::is_signed_v<T> && std::is_integral_v<T>;

/// @brief Concept for any 32-bit integer (signed or unsigned)
template <typename T>
concept Integer32 = sizeof(T) == 4 && std::is_integral_v<T>;

/// @brief Concept for 64-bit signed integers
template <typename T>
concept SignedInteger64 =
    sizeof(T) == 8 && std::is_signed_v<T> && std::is_integral_v<T>;

/// @brief Concept for 64-bit unsigned integers
template <typename T>
concept UnsignedInteger64 =
    sizeof(T) == 8 && !std::is_signed_v<T> && std::is_integral_v<T>;

/// @brief Concept for any 64-bit integer (signed or unsigned)
template <typename T>
concept Integer64 = sizeof(T) == 8 && std::is_integral_v<T>;

/// @brief Concept for any 32-bit float 
template <typename T>
concept Float32 = sizeof(T) == 4 && std::is_floating_point_v<T>;

/// @brief Concept for any 64-bit float 
template <typename T>
concept Float64= sizeof(T) == 8 && std::is_floating_point_v<T>;

/// @brief Concept for any integral type
template <typename T>
concept Integer = std::is_integral_v<T>;

// @brief Concept for streamable type
template <typename T>
concept Streamable = requires(std::ostream os, T value) {
    { os << value };
  };

// @brief Concept for string like objects
template<class T>
concept StringLike = std::is_convertible_v<T, std::string_view>;
```

```suggestion
template <Streamable Out, UnsignedInteger32 T>
static inline void write(Out& out, T x) {
  out.write(reinterpret_cast<const char*>(&y), sizeof(y));
}

template <Streamable Out, Float64 T>
static inline void write(Out& out, T x) {
  auto bytes = reinterpret_cast<const char*>(&x);
  out.write(bytes, 8);
}

template <Streamable Out, StringLike T>
static inline void write(Out& out, const T& s) {
  write(out, s.size());
  out.write(s.data(), static_cast<std::streamsize>(s.size()));
}
// Exercise for the reader
template <Streamable Out>
static void write_fixed_header(Out& out, std::size_t dim,
const std::vector<ChainRecord>& chains) {
```

SteveBronder (commented):
A few optional comments but overall looks good! I need to still look to see if there is anything else in the triple buffer code but I think this is good to merge

SteveBronder (approved):

bob-carpenter:
Thanks much @SteveBronder.  I was going to go through and make your suggested changes this week before merging.  I just need to go through a quick tutorial on concepts first.

WardBrian:
I'm not convinced concepts are necessarily a great fit here. In particular, I don't see why we'd need `write` to appear in a generic context, so overloading it is purely for style, and it's not clear that the writer code here would even persist for long enough to polish it

PR #52, opened 2026-05-12
Posterior summary

bob-carpenter:
All of the posterior summary functions from [stan/analyze/mcmc](https://github.com/stan-dev/stan/tree/develop/src/stan/analyze/mcmc) generalized to ragged chains.


include/walnuts/summary.hpp
```diff
@@ -0,0 +1,598 @@
...
+#include <stdexcept>
+#include <vector>
+
+#include <Eigen/Dense>
+#include <unsupported/Eigen/FFT>
+
+namespace walnuts {
+
+  namespace {
+    static double sample_variance(const Eigen::VectorXd& x) {
```
WardBrian: I get a warning about this function being unused

include/walnuts/summary.hpp
```diff
@@ -0,0 +1,598 @@
...
+      for (auto end : chain_ends) {
+        Eigen::Index len = static_cast<Eigen::Index>(end) - start;
+        Eigen::MatrixXd chain = draws.middleRows(start, len);
+        acov.middleRows(start, len) = autocovariance_chain(chain, fft);
+        start+= len;
+      }
+      return acov;
+    }
+
+    static Eigen::MatrixXd autocorrelation(const Eigen::MatrixXd& draws,
```
WardBrian: Also unused

examples/summary.cpp
```diff
@@ -0,0 +1,85 @@
...
+  
+  std::cout << "sample standard_deviation: " << walnuts::sample_standard_deviation(chains) << "\n\n";
+
+  Eigen::VectorXd probs(3);
+  probs << 0.05, 0.5, 0.95;
+  Eigen::MatrixXd quantiles = walnuts::quantiles(chains, probs);
+  std::cout << "quantiles(" << probs.transpose() << "):\n" << quantiles << "\n\n";
+
+  Eigen::RowVectorXd r_hat = walnuts::r_hat(chains);
+  std::cout << "R-hat = " << r_hat << "\n\n";
```
WardBrian, line 50: Running the summary example on my machine crashes after printing the quantiles:

```
quantiles(0.05  0.5 0.95):
3.18409 5.47684  8.7762 10.6697
4.05274 9.39437   15.97 25.3516
4.89986 12.6908 23.7049 37.2532

summary-example: /home/brian/Dev/cpp/walnuts/build/_deps/eigen/Eigen/src/Core/CwiseBinaryOp.h:116: Eigen::CwiseBinaryOp<BinaryOp, Lhs, Rhs>::CwiseBinaryOp(const Lhs&, const Rhs&, const BinaryOp&) [with BinaryOp = Eigen::internal::scalar_quotient_op<double, double>; LhsType = const Eigen::ArrayWrapper<Eigen::Matrix<double, 1, -1> >; RhsType = const Eigen::ArrayWrapper<Eigen::Matrix<double, -1, 1> >; Lhs = Eigen::ArrayWrapper<Eigen::Matrix<double, 1, -1> >; Rhs = Eigen::ArrayWrapper<Eigen::Matrix<double, -1, 1> >]: Assertion `aLhs.rows() == aRhs.rows() && aLhs.cols() == aRhs.cols()' failed.
Aborted (core dumped)

```

include/walnuts/summary.hpp
```diff
@@ -0,0 +1,598 @@
...
+        throw std::invalid_argument("number of rows in lps must match size of chain_id");
+      }
+      if (chain_id.size() < 1) {
+        throw std::invalid_argument("require at least one draw");
+      }
+      
+      for (std::size_t n = 1; n < chain_id.size(); ++n) {
+        if (chain_id[n] < chain_id[n-1]) {
+          throw std::invalid_argument("chain IDs must be in increasing order");
+        }
```
WardBrian: This seems like an odd requirement. Is it just because you need to detect when they change?

It seems like a simpler API might be to accept a vector of chain _lengths_, with the requirement that sum(lengths) == draws.rows()

include/walnuts/summary.hpp
```diff
@@ -0,0 +1,598 @@
...
+
+    
+  }  // end namespace
+
+  /**
+   * @brief A class for representing a collection of Markov chains.
+   * 
+   * The class can be constructed from either sorted long-form data
+   * with chain identifiers for each row or a collection of matrices,
+   * one per chain.
```
WardBrian: Is this comment out of date?

I think more needs to be said about what the input format is here

include/walnuts/summary.hpp
WardBrian, line 1: These routines in general are prime candidates for writing unit tests. If we start by testing with non-ragged draws, we may even be able to compare to Stan (or `posterior` or any other existing implementation)?

include/walnuts/summary.hpp
```diff
@@ -0,0 +1,598 @@
...
+        Eigen::Index end = static_cast<Eigen::Index>(chain_ends[static_cast<std::size_t>(k)]);
+        Eigen::Index length = end - start;
+        auto chain = draws.middleRows(start, length);
+        auto chain_mean = mean(chain);
+        mu.row(k) = chain_mean;
+        sigma_sq.row(k) = sample_variance(chain, chain_mean);
+        start = end;
+      }
+      return (1.0 + sample_variance(mu).array() / mean(sigma_sq).array())
+        .sqrt().matrix();
```
WardBrian: GDB claims the crash I am getting in the example stems from this division

include/walnuts/summary.hpp
bob-carpenter: :-).  I can work on tests.  I can probably go back in GitHub and find Stan's old tests.  Current ones won't match because everything's converted to ranks.  I guess I could go one step further and convert the columns to ranks here.  Then results should match.

include/walnuts/summary.hpp
bob-carpenter: I need the equivalent of the bat signal for @SteveBronder!

The usual issue is dangling references in expression templates but there's nothing to dangle here because both `mean()` and `sample_variance()` return concrete `RowVectorXd` types.  Then the whole expression template should be evaluated to return a `RowVectorXd`.

include/walnuts/summary.hpp
bob-carpenter: Yes, the comment's out of date.  

I'm worried about not duplicating all the data.  The handler I have built in builds a `std::vector` of matrices, so maybe I should be using that as an input format?  On the other hand, it serializes them in long form with indexes as I used here.  I might be able to generalize because I'm building views of those sliced per-chain matrices here to send to the various algorithms.

include/walnuts/summary.hpp
bob-carpenter: It's more because it matches the serialization output of the handler I wrote.  It outputs in long form with an index for the chain ID.  I could instead change that to output chain lengths.  I could also rewrite all of this very easily to work on `std::vector<Eigen::MatrixXd>` rather than `(MatrixXd, vector<size_t>)`.

include/walnuts/summary.hpp
bob-carpenter: Autocorrelation is something people like to plot for posterior summaries.  This isn't the way to give that to them, though, so I removed.  I added an issue to add average autocorrelation output across chains:

https://github.com/flatironinstitute/walnuts/issues/53

include/walnuts/summary.hpp
bob-carpenter: I'll change the input here and put the work of constructing the starts on the caller.  I also think it's a good idea to make this length-based.

include/walnuts/summary.hpp
WardBrian: I only flagged it because it was an unused function in an anonymous namespace, so it wasn't even accessible to users. If you want to keep it but moved to the top-level, also fine

examples/summary.cpp
bob-carpenter: See below.

include/walnuts/summary.hpp
bob-carpenter: Belay that bat-request.  The issue is that `mean()` was returning a `VectorXd` whereas `sample_variance()` was returning `RowVectorXd`.  It shouldn't have worked on my machine.

include/walnuts/summary.hpp
bob-carpenter: I updated the comment so that it's correct.  I'll also add a constructor with lengths.

bob-carpenter:
I'm struggling with the fact that half of these functions want all the draws concatenated (mean, variance, quantiles) and the other half want them broken out by chain (autocovariance, ess, mcse).  

The handler for the sampler collects them as a vector of matrices, so they're already broken out by chain.  I  don't see how to make a unified matrix in memory without doubling the memory overhead.  My serialization writes out in long form (a single unified matrix), though it'd be easy to read back in either way.  So one way to get around the memory doubling would be to serialize, free the in-memory representation, then read back in.  But that just replaces memory pressure with network/disk pressure.  Not sure that's a good idea.

If I store chain rather than in one big matrix in `MarkovChains`, then I have to internally concatenate the columns to sort to calculate things like quantiles.  The mean and variance I can handle just walking over the pieces.  But not quantiles.  At least if I concatenate the rows together for quantiles, I only need to do it one variable at a time.

My current path is to provide two implementations, `MarkovChains1` and `MarkovChains2`.  Version 1 will store as a big matrix and version 2 will store as individual matrices.  Both will provide convenience methods to convert what's necessary.  The algorithms can all be templated to just work against the interfaces.  This isn't that hard as I have everything using matrix views from the unified store, which are easy to replace with concrete matrices.

This feels like it may be overengineering.  Maybe I just store the separated version and pay the cost of building up for quantiles?  I think that's where I'll land.

WardBrian:
> This feels like it may be overengineering. Maybe I just store the separated version and pay the cost of building up for quantiles? I think that's where I'll land.

I certainly think it makes sense to pick one internal representation and then do whatever copy/reshaping is necessary if the user supplies a different one.

As a complicating factor, I'll throw out there that it would be nice if some of these could support getting Eigen::Map types as inputs, since that would allow me to pass in raw memory from a numpy array up in Python with no copy overhead. Obviously quantiles would need to copy before sorting, but I think most of the other operations would be amenable to this?

bob-carpenter:
> it would be nice if some of these could support getting Eigen::Map types as inputs

Good point.  I'll see if I can make the reference that gets stored more generic rather than just copying.

I built everything so I can use either base representation, but I have only implemented the contiguous basis one so far.  It'll be very easy given the concept to implement a class based on one matrix of draws per chain.

bob-carpenter:
@WardBrian:  This is all revised and ready to review again.  I generalized so the chains object can be constructed with a `Map`.

include/walnuts/summary.hpp
bob-carpenter: I started a new branch where I'm going to try to go through and write unit tests for everything I can, not just `summary.hpp`.  I'm open to suggestions on how to stage this.

include/walnuts/summary.hpp
```diff
@@ -0,0 +1,786 @@
...
+
+#include <Eigen/Dense>
+#include <unsupported/Eigen/FFT>
+
+#include <walnuts/concepts.hpp>
+
+namespace walnuts {
+
+template <typename Derived>
+inline Eigen::RowVectorXd col_means(const Eigen::MatrixBase<Derived>& draws) {
```
WardBrian, line 16: Does matrixbase accept Map objects? I see these helpers getting called with the result of `chains_views` and want to make sure they're not copying unnecessarily

include/walnuts/concepts.hpp
```diff
@@ -259,3 +259,34 @@ concept AdaptiveSampler = requires(A& a, const A& ca) {
...
+ *  - `min_chain_size()` returning a value convertible to
+ *    `Eigen::Index`, the length of the shortest chain.
+ *  - `chain_view(std::size_t)` returning a type convertible to
+ *    `Eigen::MatrixXd`, a view of one chain (one draw per row).
+ *  - `draws(Eigen::Index)` returning a type convertible to
+ *    `Eigen::VectorXd`, the sequence of draws for one variable across
+ *    all chains.
+ */
+template <typename M>
+concept MarkovChainSequence =
```
WardBrian, line 284: nit: if this is only used in the signatures of internal functions I wonder if it should just live in the summary file?

include/walnuts/summary.hpp
```diff
@@ -0,0 +1,786 @@
+#pragma once
+
+#include <algorithm>
+#include <complex>
```
WardBrian: I believe the complex header is not needed here

WardBrian (approved):
A couple small things but overall this looks good. For the tests you're writing, I think whatever form of PR you find easiest is fine

include/walnuts/summary.hpp
bob-carpenter: Yes.   `Eigen::Map` inherits from `Eigen::MatrixBase`.  It's very hard to figure that out from their code, but all the meta-doc agrees.  It also agrees that I should just be using `Eigen::Ref` here if I mainly care about `Map` and `MatrixXd` cases.

include/walnuts/concepts.hpp
bob-carpenter: I was torn on this.  Putting all the concepts in one place seemed easier, but I'm flexible.  I now have two different `MarkovChainSequence` implementations, so it seemed worth pulling this out as the way to relate them.

include/walnuts/summary.hpp
bob-carpenter: <del>I think you're right.</del>  I'm bringing in complex types, but never referring to any of the types explicitly.

Update:  confirmed it's not necessary and removed.

include/walnuts/concepts.hpp
bob-carpenter: In any case, we can move them around later if we want---I'll just leave here for now.

PR #56, opened 2026-05-15
Remove duplicated passing of initial position

WardBrian:
Closes #54

PR #58, opened 2026-05-15
Update Stan CLI for new configs 

WardBrian:
Builds on #56 

This does not currently connect to the auto-warmup/auto-stopping code in api.hpp, all I've done is restore the previous cli to a compiling state. Integration with the api.hpp code to follow

### DIFF
--- a/examples/stan_cli.cpp
+++ b/examples/stan_cli.cpp
@@ -106,14 +106,11 @@ class StanHandler {
 };
 
 template <typename RNG>
-StanHandler run_walnuts(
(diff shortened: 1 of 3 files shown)


WardBrian:
cc @bgoodri -- you mentioned wanting to throw walnuts at a bridgestan model -- if you do so in the next few days, this is the branch you'll want

bob-carpenter (approved):
Thanks so much for updating all this---it's what anyone coming here will try to use until we have R and Python interfaces.

Issue #57, opened 2026-05-15
SamplingConfig's min_micro_steps is never used


WardBrian:
In all the places it would be used, we end up using the value from the MinMicroStepsAdaptHandler instead

bob-carpenter:
Fixed with PR #60

PR #60, opened 2026-05-17
plumb min micro; fixes issue #57

bob-carpenter:
This plugs in the min-micro steps by making sure there are at least that many steps with a max(min-micro, X) number of micro steps, where X is the best estimate from the algorithm for the target.  This could potentially set up a conflict in that the user supplies a min micro where we can't achieve their target macro steps.


bob-carpenter:
@WardBrian and @SteveBronder:  I'm just pushing the simple fixes, but feel free to take a look at this and the last tiny PR I did.

PR #62, opened 2026-05-20
Details refactor

bob-carpenter:
There are several things in this PR:

* moving all the non-public-API functions to `walnuts::details`
* finishing missing doc, adding @brief so summaries would show up
* reorganizing and factoring `examples/CMakeLists.txt`
* moving iwyu, etc. to cmake directory and updating README.md
* thread sanitizer suppression (apparently, `TSan` cannot handle the `TripleBuffer` pattern because it can't establish that the reader and writer never access the same slot simultaneously)
* expand example sizes so they don't finish before monitoring without `TSan` turned on

So you don't have to build the doc yourselves, here's what I've put into `walnuts` namespace---everything else is in `walnuts::detail`.

```
PRIMARY
------------------------------------------------------------ 
Imports:
* <walnuts>

<walnuts/api>
Concepts  
* concept LogpGrad                      (concepts.hpp)  
* concept ChainHandler                  (concepts.hpp)  
* concept GlobalHandler                 (concepts.hpp)  
* concept IntreruptCallback             (concepts.hpp)
* concept MarkovChainSequence           (concepts.hpp)

Classes
* CppInterruptCallback                  (handlers.hpp)
* GlobalStore                           (handlers.hpp)
* ChainStore                            (handlers.hpp)
* InitChainConfig                       (config.hpp) 
* InitConfig                            (config.hpp) 
* InitConfigBuilder                     (config.hpp) 
* WarmupConfig                          (config.hpp) 
* WarmupConfigBuilder                   (config.hpp) 
* SamplingConfig                        (config.hpp) 
* SamplingConfigBuilder                 (config.hpp) 
* WalnutsConfig                         (config.hpp) 
* MarkovChainsSplit
* MarkovChainsUnified

Functions
* walnuts()                             (api.hpp)
* operator<<(, InitConfig)              (config.hpp) 
* operator<<(, WarmupConfig)            (config.hpp) 
* operator<<(, WarmupConfig)            (config.hpp)
* write_step_size                       (handlers.hpp)
* write_mass_matrix                     (handlers.hpp)
* write_sample                          (handlers.hpp)
* mean                                  (summary.hpp)
* sample_variance                       (summary.hpp)
* sample_standard_deviation             (summary.hpp)
* quantiles                             (summary.hpp)
* autocovariance                        (summary.hpp)
* r_hat                                 (summary.hpp)
* effective_sample_size                 (summary.hpp)
* monte_carlo_standard_error            (summary.hpp)

SECONDARY
------------------------------------------------------------
Concepts
* SampleHandler                         (walnuts.hpp)

Classes
* WalnutsSampler                        (walnuts.hpp)
* AdaptiveWalnuts                       (adaptive_walnuts.hpp)
* Random 
```


include/walnuts/adaptive_walnuts.hpp
```diff
@@ -168,6 +166,11 @@ class MinMicroStepsAdaptHandler {
   double count_;
 };
 
+}  // namespace detail
+}  // namespace walnuts
+
+namespace walnuts {
```
WardBrian, line 172: Is there a reason to close and then re-open the same namespace?

include/walnuts/walnuts.hpp
```diff
@@ -630,12 +634,12 @@ class WalnutsSampler {
    * @throw std::invalid_argument If `max_error` is not positive or not finite.
 
    */
-  WalnutsSampler(Random<RNG>& rand, H& sample_handler, const F& logp_grad,
+  WalnutsSampler(RNG& rng, H& sample_handler, const F& logp_grad,
```
WardBrian, line 637: I understand the desire to leave Rand out of the public API, I just want to note that the std prongs will be much harder to mock than a Rand does

cmake/tsan-suppressions.txt
WardBrian, line 1: This makes me pretty nervous, because the triple buffer is the main place we'd like TSan to actually catch issues. The suppressions are really only supposed to be used when you've also opened a bug report on the defect, and my prior is that it is significantly more likely we have an actual issue in the triple buffer than it is that TSan is incorrect.

examples/CMakeLists.txt
```diff
@@ -1,26 +1,45 @@
...
+# Apply the standard example flags (and TSan if enabled) to a target,
+# plus optionally register a `tsan_<target>` runner with suppressions.
+function(walnuts_add_example name)
+  target_compile_options(${name} PRIVATE -O3 -Wall)
+  if(WALNUTS_USE_TSAN)
+    target_compile_options(${name} PRIVATE -fsanitize=thread)
+    target_link_options(${name}    PRIVATE -fsanitize=thread)
+    add_custom_target(tsan_${name}
+      COMMAND ${CMAKE_COMMAND} -E env
+        "TSAN_OPTIONS=suppressions=${CMAKE_SOURCE_DIR}/docs/tsan-suppressions.txt"
```
WardBrian: Though, it also looks like this is the wrong path, so maybe the supressions aren't actually doing anything?

examples/CMakeLists.txt
```diff
@@ -1,26 +1,45 @@
...
     GIT_REPOSITORY https://github.com/CLIUtils/CLI11.git
-    GIT_TAG v2.5.0
+    GIT_TAG        v2.5.0
   )
   FetchContent_MakeAvailable(cli11)
+
   add_executable(stan_cli stan_cli.cpp)
-  target_link_libraries(stan_cli PRIVATE Eigen3::Eigen walnuts BridgeStan::BridgeStan CLI11::CLI11)
-  target_compile_options(stan_cli PRIVATE -Wall)
+  target_link_libraries(stan_cli PRIVATE walnuts BridgeStan::BridgeStan CLI11::CLI11)
```
WardBrian: Even though walnuts will include Eigen, I like having it here as a sort of documentation, because the cli does actually name Eigen types directly

examples/CMakeLists.txt
```diff
@@ -1,26 +1,45 @@
...
+  target_compile_options(stan_cli PRIVATE -Wall)  # no -O3, no TSan — matches your current file
```
WardBrian: Comment seems odd

include/walnuts/handlers.hpp
WardBrian, line 1: I still think we should consider excluding this from the public api, either by moving all of it to `detail` or just moving the file to the examples folder

WardBrian (commented):
A few minor comments. The only thing I'd consider a blocker is the TSan items -- I don't think suppressing the warning is sufficiently justified at the moment

include/walnuts/adaptive_walnuts.hpp
bob-carpenter: It makes it easier to do a reverse search for "namespace" and figure out which namespace you're in.  I didn't make this style up, but I'm flexible.

include/walnuts/walnuts.hpp
bob-carpenter: That's a good point.  As we discussed, I may have to rethink this when it comes to testing.  On the other hand, if we're only testing top-level behavior, then sufficient randomness and a wide set of tests should cover all the non-exception branches of the code.

cmake/tsan-suppressions.txt
bob-carpenter: Indeed.  I'll replace with a mutex-based version that doesn't throw warnings, but isn't lock free, then we can revisit going back to a triple buffer later, perhaps with @SteveBronder's help.

examples/CMakeLists.txt
bob-carpenter: I'll just remove the suppressions for now.

examples/CMakeLists.txt
bob-carpenter: OK, I put that back.

include/walnuts/adaptive_walnuts.hpp
WardBrian: If you're going to do it consistently, you might as well do `namespace walnuts::detail { ... }` for the detail namespace to save a line.

examples/CMakeLists.txt
bob-carpenter: Also, clearly AI generated, right down to the contrastive style, even in comments.  I don't know who decided that all the LLMs were all about contrastive sentences all the time.

include/walnuts/handlers.hpp
bob-carpenter: I think it's worth providing the handlers that just collect all the output and provide a standard I/O format.

PR #64, opened 2026-05-25
Unit tests

bob-carpenter:
This adds the easy batch of unit tests for functions that are easily unit testable. 

1.  `test_summary.cpp`:  Tests all functions in `walnuts/include/summary.hpp`.  100% code coverage (function, line, region, and branch).
2.  `test_config.cpp`:  Tests all functions in `walnuts/include/config.hpp`.  100% code coverage (function, line, region, and branch).

Claude Sonnet 4.6 was super helpful.  I found Opus 4.7 to be overthinking it.  I cleaned up and introduced bad example functions to generalize tests.  

100% code coverage in summary revealed two pieces of dead code in the original.  Both were testing conditions that could never arise because of the `- 4` offset in the end condition.  The offset came from Stan's implementation and it ensures you don't estimate correlations from even fewer examples than that. I also had to be careful testing both the split and unified representations of Markov chains because the coverage tool was very picky about that.


include/walnuts/config.hpp
```diff
@@ -1045,7 +1045,55 @@ inline std::ostream& operator<<(std::ostream& out, const SamplingConfig& cfg) {
...
  */
-struct WalnutsConfig {
+class WalnutsConfig {
+public:
+
+  /**
+   * @brief Construct a Walnuts configuration given the component
+   * configurations.
+   *
+   * The arguments will be moved if they are rvalues and copied if lvalues.
```
WardBrian, line 1055: I think this comment is incorrect -- they are taken by value, which means they will always have at least one copy performed

WardBrian (commented):
I didn't look too closely since it was already merged, but one thing did stand out to me

include/walnuts/config.hpp
bob-carpenter: Thanks for taking a look.   I didn't think there'd be much in this one that would be a problem, but the triple buffer code could use a look.  I just didn't want to chain a whole sequence of branches because I thought I'd get confused.  

For this particular case, I thought the argument would be move constructed in the constructor, then move constructed in the class.  Am I missing a move constructor for the object being passed?

WardBrian:
The following tests are failing on my machine:

         51 - Autocovariance.LagZeroEqualsChainBiasedVariance (Failed)
         52 - Autocovariance.FullResultMatchesReference (Failed)
         53 - Autocovariance.MatchesDirectComputationPerChain (Failed)
         55 - Autocovariance.BothTypesAgreeOnSameData (Failed)
         56 - Autocovariance.SingleDrawChainGivesZero (Failed)
         75 - EffectiveSampleSize.BothTypesAgreeOnSameData (Failed)
         79 - MonteCarloStandardError.EqualsStdDevOverSqrtEss (Failed)
         82 - MonteCarloStandardError.BothTypesAgreeOnSameData (Failed)


Most of the errors are rather uninformative because of the use of EXPECT_TRUE, e.g.:

```
[ RUN      ] MonteCarloStandardError.EqualsStdDevOverSqrtEss
/home/brian/Dev/cpp/walnuts/tests/summary_test.cpp:1179: Failure
Value of: mcse.isApprox((sd.array() / ess.array().sqrt()).matrix(), 1e-10)
  Actual: false
Expected: true

```

include/walnuts/config.hpp
WardBrian: The classes you accept here have `= default`'d constructors, so they should have move constructors, but those are in a private block, which I'm not sure if it effects or not. It's probably fine either way

bob-carpenter:
I'll start a new PR and write my own `VectorXd`/`MatrixXd` comparison code and replace the `isApprox()` calls from Eigen.   Because of how noisy ess estimates are, I'm guessing the one you see there is just arithmetic error.  I wouldn't expect 10 digits of precision out of all these functions composed together.

I'll also go through and make sure the move constructors are public where they need to be.

include/walnuts/config.hpp
bob-carpenter: The relevant copy and move constructors are public.  If they weren't, this wouldn't compile because the move constructors wouldn't be available.  Only the nullary constructors are declared to take their default value in the private block because I only want versions to be constructible through the builder.

bob-carpenter:
Is there any way we can recreate some of these tests with CI so that I can see them?  Or can you just give me the dump you have and maybe I can fix some of the issues?

I could probably configure gcc to run on my Mac if you think that's the source of the problem.

Otherwise, maybe we can sit down tomorrow or some time this week and debug on your machine.  

I created a new branch, `more-unit-tests`, where I went and replaced all the `.isApprox()` with a locally defined `expect_near` for vectors.  I'm curious if that gives you more information about what's failing and by how much.   Or if there are other tests I need to instrument more finely.

WardBrian:
See #65 for them failing in on Windows/Linux CI. You can push straight to that branch if you’d like

PR #68, opened 2026-05-28
Silence warning for possibly-uninitialized variable

WardBrian:
My compiler was complaining about these out parameters of the `macro_step` function, since it couldn't be certain that the supplied `logp_grad` actually set anything before their first usage


bob-carpenter (changes requested):
Should we really do this or should we just silence the warning?  If we go with this assignment, we should include a comment that the values are meaningless and always get overwritten by the macro step.

PR #70, opened 2026-06-02
Test util

bob-carpenter:
This is a pretty comprehensive test for everything in `walnuts` (not `walnuts::detail`) for `summary.hpp`.  

I also reformatted, so there's a bunch of distractions from that---sorry about that.

### DIFF
--- a/tests/util_test.cpp
+++ b/tests/util_test.cpp
@@ -5,8 +5,8 @@
 #include <stdexcept>
 #include <vector>
 
-#include <../tests/test_util.hpp>
 #include <walnuts.hpp>
+#include "test_util.hpp"
 
 // Random class *****************************************************
 
--- a/tests/util_test.cpp
+++ b/tests/util_test.cpp
@@ -155,27 +155,21 @@ TEST(LogSumExp, NaNPropagates) {
   EXPECT_TRUE(std::isnan(walnuts::detail::log_sum_exp(nan, nan)));
 }
 
-// walnuts::detail::log_sum_exp(VectorXd) function ***********************************
+// walnuts::detail::log_sum_exp(VectorXd) function ******************
 
 TEST(LogSumExpVector, EmptyVector) {
   Eigen::VectorXd x(0);
   EXPECT_EQ(walnuts::detail::log_sum_exp(x), -std::numeric_limits<double>::infinity());
 }
 
-TEST(LogSumExpVector, MatchesTwoArgVersion) {
-  Eigen::VectorXd x(2);
-  x << 1.0, 2.0;
-  EXPECT_NEAR(walnuts::detail::log_sum_exp(x), walnuts::detail::log_sum_exp(1.0, 2.0), 1e-15);
-}
-
-// Single element: log(exp(x)) = x
 TEST(LogSumExpVector, SingleElement) {
-  Eigen::VectorXd x(1);
-  x << 3.7;
-  EXPECT_DOUBLE_EQ(walnuts::detail::log_sum_exp(x), 3.7);
+  for (auto x : std::vector<double>{-1000.0, 0.0, 3.7, 1000.2}) {
+    Eigen::VectorXd v(1);
+    v << x;
+    EXPECT_DOUBLE_EQ(walnuts::detail::log_sum_exp(v), x);
+  }
 }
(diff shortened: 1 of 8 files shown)


tests/util_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+TEST(Detail, FalseSharingGuardSizeIs128) {
+  EXPECT_EQ(walnuts::detail::FALSE_SHARING_GUARD_SIZE, std::size_t{128});
+}
+
+// Update enums
+
+TEST(Detail, UpdateEnumValuesAreDistinct) {
+  EXPECT_NE(walnuts::detail::Update::Barker,
+            walnuts::detail::Update::Metropolis);
+}
```
WardBrian: A couple thoughts:

First, this could be a static assert, no need to actually run a test at runtime.

Secondly, I think this (and all of the other 'AreDistinct') are bad tests. The only way they'd ever fail is if someone manually set the values in the enums to the same constant, which would be trivial to catch on code review. But, by naming them in a test, we still incur the extra pain if we ever wanted to rename them or similar.

tests/util_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include <../tests/test_util.hpp>
+#include <walnuts.hpp>
+
+// Constants ********************************************************
+
+// FALSE_SHARING_GUARD_SIZE constant
+
+TEST(Detail, FalseSharingGuardSizeIs128) {
+  EXPECT_EQ(walnuts::detail::FALSE_SHARING_GUARD_SIZE, std::size_t{128});
```
WardBrian: Similarly, the only way this test fails is if someone changes the constant, in which case they'd probably also just come here and update the test, so I don't see any value add

tests/util_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+// Direction enums
+
+TEST(Detail, DirectionEnumValuesAreDistinct) {
+  EXPECT_NE(walnuts::detail::Direction::Forward,
+            walnuts::detail::Direction::Backward);
+}
+
+// Direction types
+
+TEST(Detail, ForwardAndBackwardTypesCarryCorrectValues) {
```
WardBrian: This is just testing that the language spec is correct, really

WardBrian (changes requested):

tests/util_test.cpp
bob-carpenter: I went with the "secondly" and removed them.

tests/util_test.cpp
bob-carpenter: Removed.

tests/util_test.cpp
bob-carpenter: And that we didn't reverse them, but I'm OK removing and did.

bob-carpenter:
I realized I didn't finish these before submitting the PR.  Let me go back and do that before you merge.

bob-carpenter:
@WardBrian  I added the rest of the tests, so this is ready to review again.

tests/util_test.cpp
```diff
@@ -0,0 +1,384 @@
+#include <gtest/gtest.h>
+
+#include <Eigen/Dense>
+#include <cmath>
+#include <stdexcept>
+#include <vector>
+
+#include <../tests/test_util.hpp>
```
WardBrian: .. in a `#include` is an odd pattern (I hadn't noticed us doing it in the other tests). `#include "test_util.hpp"` would be more like what I'd expect

WardBrian (approved):
One minor thing but otherwise looks good to me

tests/util_test.cpp
bob-carpenter: I added 

```cmake
target_include_directories(${TEST_NAME}_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
```

 into the `add_test()` function in `tests/CMakeLists.txt` and changed includes to

```cpp
#include "test_util.hpp"
```

WardBrian: Adding it as an include directory is unnecessary, since all implementations I am aware of follow the convention that `#include "..."` searches the parent directory of the current file first. But it doesn't hurt either way

PR #77, opened 2026-06-12
unroll leapfrogs; fixes #30

### LINKED ISSUE #30: unfold leapfrog loop to avoid double rho updates
We currently have the naive leapfrog implementation:

```cpp
    for (Integer n = 0; n < num_steps; ++n) {
      rho_next.noalias() += half_step * grad_next;
      theta_next.noalias() +=
          step * (inv_mass.array() * rho_next.array()).matrix();
      logp_grad(theta_next, logp_next, grad_next);
      rho_next.noalias() += half_step * grad_next;
    }
```

This should be unrolled to avoid the double update to `rho_next`.

bob-carpenter:
This just unrolls the two implementations of leapfrog within `walnuts.hpp` (the regular one and the one to check reversibility).

### DIFF
--- a/include/walnuts/walnuts.hpp
+++ b/include/walnuts/walnuts.hpp
@@ -240,7 +240,7 @@ static bool within_tolerance(const F& logp_grad,
   rho_next += half_step * grad_next;
   logp_next += logp_momentum(rho_next, inv_mass);
   return std::abs(logp_next - logp) <= max_error;  // only tests one way
-}  
+}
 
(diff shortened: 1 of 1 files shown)


WardBrian:
@bob-carpenter not sure if you noticed that this never got merged, there's now a conflict that would need to be resolved

bob-carpenter:
I can fix later.  It's a minor performance enhancement

PR #78, opened 2026-06-15
Python interface

### LINKED ISSUE #1: Python interface
_Copying from https://github.com/bob-carpenter/walnuts/issues/38_

#### Goal

A lightweight Python interface for 

* configuring a sampler including defaults and error checking/reporting
* sampling from a target density function using NUTS or WALNUTS
* providing the draws and metadata in standardized formats

#### Target

A target log density needs to specify all of the following.

* constraining transfrom to dictionary or flat vector (?) (for output)
    * run an RNG component like Stan's generated quantities after draws rather than built in requirement
* unconstraining transform from dictionary (for initialization)
* unconstrained log density function
* gradient of unconstrained log density function
* dimensionality

This does *not* require a generated quantities step like in Stan's `generated quantities` block.  That can be done in postprocessing given the draws directly in Python.

Ideally we'd like to make it easy to adapt targets from the following systems to start.

* Stan
* PyMC
* JAX
* plain old Python

#### Output

The output needs to include:

* parameter names and shapes (on constrained scale)
* draws (on constrained scale)
* diagnostics: log density, gradient norm, divergences, gradient evals
* adaptation: mass matrix, (macro) step size
* warmup iterations should be optional

It would be nice to have a feature like CmdStan's `stan_variable` to get structured output.

The overall output should be easy to export to [`ArviZ`](https://github.com/arviz-devs/arviz), but should not include a dependency if it can be avoided.

#### User-facing documentation

The documentation will be like a small subset of CmdStanPy's.  

#### Prior art

* [`nutpie`](https://github.com/pymc-devs/nutpie)
* [`cmdstanpy`](https://github.com/stan-dev/cmdstanpy)
* [`tinystan`](https://github.com/WardBrian/tinystan)

WardBrian:
This PR starts us down the path to a Python library that can run the Walnuts sampler.

Implemented:
- [x] Simple, flat C API for BridgeStan models
- [x] Error and interrupt handling (both cannibalized from [`tinystan`](https://github.com/WardBrian/tinystan))
- [x] Basic Python wrapper

   In particular, the following works if run from the `walnuts` top-level directory:
   `pip install .`
   ```python
   import bridgestan
   from walnuts import walnuts_stan # temporary import name
   model = bridgestan.StanModel(...)
   draws = walnuts_stan(model)
   ```
- [x] This PR depends on [currently-unreleased functionality in BridgeStan](https://github.com/roualdes/bridgestan/pull/325)
- [x] Support for pymc models and generic python functions
- [x] Packaging/build for this new code. In particular, will have to decide what to do about me including `../examples/load_stan.hpp`
- [x] Docs


WardBrian:
Latest push has the code that works for a generic python function `logp(theta) -> (lp, grad)`, and a fast-path for numba users. 

Using the code in `ex.py` (`ipython -i ex.py`):
```
In [1]: %timeit walnuts_stan(m)
1.94 ms ± 38.8 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [2]: %timeit walnuts_pyfunc(logp, num_params=2)
1.33 s ± 336 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit walnuts_pyfunc(logp_numba, num_params=2)
1.65 ms ± 9.71 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
So far as I can tell, that overhead is _not_ just from GIL issues, but who is to say...

The type I'm accepting for the callback is the same as nutpie uses for their pymc support, so it should easy to adapt, but I'd ask @aseyboldt for some pointers (pardon the pun)

WardBrian:
I realized there was an even simpler way to measure the Python overhead, which is to call Bridgestan _through_ the interface intended for Python functions:

```python
In [1]: %timeit walnuts_stan(m)
1.95 ms ± 38.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [2]: %timeit walnuts_pyfunc(m.log_density_gradient, num_params=2)
234 ms ± 17.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

So [two Python if statements](https://github.com/roualdes/bridgestan/blob/main/python/bridgestan/model.py#L685) and a ctypes call to the same underlying function is ~100x slower than just calling that function in the compiled code.

bob-carpenter:
@WardBrian's instructions were too terse for a Python package dev beginner like me.  The missing instructions are as follows.

## Step 1:  Install latest bridgestan.  

This can be done anywhere, but the export should be to the current path.  You should use your Python here.  You can do this all in a virtual environment if your Python install is too locked down to `pip install`.

```sh
cd ~/
git clone --recurse-submodules https://github.com/roualdes/bridgestan.git ~/bridgestan
python3 -m pip install ~/bridgestan/python
export BRIDGESTAN=~/bridgestan
```

## Step 2:  Compile with `STAN_THREADS=true`

Within Python, `bridgestan` needs to be called so that it uses `STAN_THREADS=True` or it will segfault when trying to run multiple chains in parallel.   The following should go into the top-level file `walnuts/example.py`:

```python
import numpy as np
import bridgestan
from python import walnuts_stan

model = bridgestan.StanModel(
    "simple.stan",                      # path to your .stan; compiled automatically
    make_args=["STAN_THREADS=True"],    # chains run on threads
)
draws = walnuts_stan(model)

all_draws = np.concatenate([chain["alpha"] for chain in draws])
print("Simulated with mean 2.3, standard deviation 1.7")
print(f"  posterior mean = {all_draws.mean():5.2f}")
print(f"  posterior standard deviation = {all_draws.std():5.2f}")
```

## Step 3:  Run it

That worked for me and produced this output:

```sh
$ cd walnuts
$ python3 example.py
Simulated with mean 2.3, standard deviation 1.7
  posterior mean =  2.30
  posterior standard deviation =  1.62
```

WardBrian:
@bob-carpenter sorry for the missing details -- I assumed you already had bridgestan (I suppose you probably do/did in _some_ Python env, but that's not saying much!)

The latest push wires up cmake and lets you `pip install .` in the top level of the repo to build and install a package called `walnutpie` (so your example should `from walnutpie import ...` now).

Note that at the moment, it seems `pip install -e` does not work with this build set up. I spent a good chunk of yesterday trying, but nothing yet

bob-carpenter:
Thanks, that sounds great.  I don't even know what `-e` does with pip, so unless I look it up, I won't know what I'm missing.

I'll ask Adrian about putting "pie" in the name, but for now, can we just call the package "walnuts"?

WardBrian:
`-e` is short for "editable", which lets you install a package but be able to edit the source and have the changes reflected in your next `import` without re-running `pip install`. It's incredibly useful for development, but as you can imagine interacts in complicated ways with not-pure-Python packages

WardBrian:
I've renamed the python import to `walnuts` -- the name `walnutpie` is still used internally because the compiled module does need to have a distinct name than the top-level header-only library for build purposes

WardBrian:
Wheels are now building and sucessfully importing on all the major platforms except Intel Macs (a C++20 build issue it looks like) and Windows ARM (building fine, fails to import -- not sure how important this one is to figure out)

WardBrian:
Okay, I've resolved all of my internal TODOs that needed to be addressed before this was feature-complete. Testing and doc are both needed before a release, but I'm happy with how the code looks now.

bob-carpenter:
This is great news.  I'm working on a patch to the underlying code that will plug in the NUTS approach to initializing step size. That's one of the upshots of the Discourse discussion of a model that Walnuts is struggling with.

WardBrian:
My latest push includes the start of some tests up at the python level. At the moment they are quite basic, just testing that the requested number of draws gets returned and that the same seed does reproduce the same (prefix-of) draws. 

These were enough to find the deadlock from #81, but we will certainly want more

python/src/walnutpie/pyfunc.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Parameters
+    ----------
+    logp : Union[ Callable[[np.ndarray], tuple[float, np.ndarray]], &quot;numba.core.ccallback.CFunc&quot;, tuple[ctypes.CFUNCTYPE, Any], ]
+        _description_
+    num_params : Optional[int], optional
+        _description_, by default None
+    inits : Optional[np.ndarray], optional
+        _description_, by default None
+    num_chains : int, optional
```
WardBrian, line 102: @bob-carpenter this and the matching comment in python/src/walnuts/stan.py are a good place to start with doc if you'd like!

WardBrian:
@bob-carpenter this branch is now pushing to https://flatironinstitute.github.io/walnuts/latest/

bob-carpenter:
I am reading through the Python interface and I don't understand it all, and in particular, I have zero experience with the foreign function interfaces. 

1.  In `export.h`, the same visibility is set in both branches of the first conditional.

```cpp
#    ifdef walnutpy_EXPORTS
        /* We are building this library */
#      define WALNUTPY_EXPORT __attribute__((visibility("default")))
#    else
        /* We are using this library */
#      define WALNUTPY_EXPORT __attribute__((visibility("default")))
#    endif
```

2. For "# TODO: does a faster path exist for JAX?" in `pyfunc.py`, I asked Claude and it suggested that the way `pyfunc` would work for JAX is suboptimal as it pays the dispatch plus global interpreter lock (GIL) overhead.  It's suggesting two alternatives:
    - BEST: `jax.jit(jax.value_and_grad(f)).lower(x).compile()` to get a PJRT (Pretty much just another runtime)-based C interface, but it still needs to go through PJRT (the C interface) and it looks like it's still going to need to copy into the PJRT buffer format and requires some low-level C/C++ fiddling.
    - BETTER THAN CURRENT: Remove the JIT overhead from the trampoline using `.lower(x).compile()`, which is still paying the GIL cost.  
    
3.  This is duplicating some of the work of the config objects in C++ (but cutting down on contact points) in `pyfunc.py` because it's doing things like duplicating a single step size or inverse metric for all chains.  This is duplicated again in `stan.py`, so it may be useful to pull all this out into a function that can be called in `pyfunc.py` and `stan.py`.

4.  I couldn't really follow the `StanOutputs` doc as I don't know what ":meth:`~StanOutput.get`" is.  Does this give us the same kind of structured output as we get from CmdStanPy?  That's what people are really going to want.    

5.  I was confused about this in the previous documentation:

```py
    ----------
    model : bridgestan.StanModel
        The BridgeStan model to fit
```

Is this the `.so` file produced by BridgeStan as it was for the CLI?  If so, I think it'd help to say that.

6.  In the output processing, should we be preallocating the list?  It'll be small (order 4 to 64, say, with 4 being much more common), and I'm assuming it's only copying pointers, not deep copying.  So maybe not worth preallocating.

```py
outputs: list[StanOutput] = [None] * num_chains
...
    outputs[n] = output_chain
```

7.  The Python doc is not consistent about whether there's a period after parameter declarations (they're usually not full sentences, but the C++ standard is to include the periods and I tried to be consistent).  So I may have introduced a bunch of the inconsistency by including periods.  We should pick a convention and stick to it.

8.  I have zero experience with the interfaces in `handlers.hpp`.  The C++ looks fine, but I have no idea if there are better or worse ways to do this for Python.

WardBrian:
> In export.h

This file is [generated by cmake](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html), which is why it is gitignored

> jax.jit(jax.value_and_grad(f)).lower(x).compile() to get a PJRT (Pretty much just another runtime)-based C interface, but it still needs to go through PJRT (the C interface) and it looks like it's still going to need to copy into the PJRT buffer format and requires some low-level C/C++ fiddling

I believe the C API for PJRT is only available for accelerators at the moment -- there is an open ticket to add it for CPU https://github.com/jax-ml/jax/issues/37033

> This is duplicated again in stan.py, so it may be useful to pull all this out into a function that can be called in pyfunc.py and stan.py

I cleaned some of this up into functions. Some duplication is unavoidable I think

> Does this give us the same kind of structured output as we get from CmdStanPy? That's what people are really going to want.

Yes. If `theta` is a `matrix[3,4]` in the stan program, `StanOutput.get("theta")` will be a draws by 3 by 4 ndarray

> Is this the .so file produced by BridgeStan as it was for the CLI? If so, I think it'd help to say that.

No, it's the `bridgestan.StanModel` class in Python. If what you have is the path to the `.so`, you need to call that classes' constructor and also supply data first

> In the output processing, should we be preallocating the list?

I think not worth it. In particular, Python linters will complain about assigning a list of `None`s to a list of `StanOutput`s, so you'd need to opt out of typechecking for that list

> The Python doc is not consistent about whether there's a period after parameter declarations

I tend to follow the style of not using a trailing period when it is a single sentence/fragment, but using it when there are multiple sentences. I don't think the code as-is is following this, either, so I'm happy to just add them everywhere

bob-carpenter:
That's much better answers than I expected, especially w.r.t. BridgeStan and the analog of `stan_variables()` in CmdStanPy.  And thanks for the consistency pass on the doc---that kind of stuff annoys me more than it should

aseyboldt:
About calling Jax from something else than python to avoid the Gil and
python overhead: I opened this a while ago:
https://github.com/jax-ml/jax/discussions/37363
This allows calling the compiled Jax function directly if it is running on
cuda. For CPU it is still not exposed.
I'm not yet making use of it in nutpie, it does require a bit of setup work.

WardBrian:
@aseyboldt yes, I think the CPU-specific version of that request is in https://github.com/jax-ml/jax/issues/37033. We could write a small shim around the header if they did provide a shared library for it, which would be nice (and avoid us having to actually depend on jax directly in our build)

aseyboldt:
Jax is a big and hard to understand thing, but I think they are talking
about something slightly different.
`export` serialises the function, so that a different xla client can load
and execute it. I don't think that's desirable for us. It means that
nutpie/walnutpie would have to create a second xla client. At least for GPU
execution that's not a good idea, they would each create a cuda context and
fight over the memory and access of the GPU.
I think what we want is to be able to submit tasks to the already existing
XLA client that also executes all the normal Jax functions.
Pjrt defines a c API for submitting tasks, but we need to get a pointer to
the internal xla client that Jax uses.

PR #82, opened 2026-07-09
Unify patterns between adaptation threaded code and sampling threaded code

WardBrian:
I noticed while debugging #81 that sampler.hpp and adapt.hpp, despite being conceptually similar, had a ton of incidental differences in coding style that made it hard to transfer knowledge between the two files. 

The goal of this PR is just to make the two files look more like siblings; as a side effect, the adapt controller loop also now does a bunch less work, since it won't check anything until the minimum number of iterations is reached (the sampler already did this)

### DIFF
--- a/include/walnuts/adapt.hpp
+++ b/include/walnuts/adapt.hpp
@@ -1,17 +1,16 @@
 #pragma once
 
-#include <Eigen/Dense>
-#include <algorithm>
 #include <cmath>
 #include <cstddef>
 #include <deque>
-#include <exception>
 #include <functional>
 #include <latch>
 #include <limits>
 #include <stop_token>
 #include <thread>
 #include <vector>
+
+#include <Eigen/Dense>
 
 #include "walnuts/concepts.hpp"
 #include "walnuts/config.hpp"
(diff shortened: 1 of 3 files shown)


include/walnuts/adapt.hpp
```diff
@@ -91,23 +82,18 @@ class AdaptWorker {
...
    *
-   * @param[in] chain_id The identifier for which chain this is.
-   * @param[in] init_cfg The initialization configuragation.
    * @param[in] warmup_cfg The configuration for warmup.
-   * @param[in] buffer The buffer to hold the adaptation states.
-   * @param[in] start_gate A latch to gate work to start synchronously across
-   * workers.
    * @param[in] adapter The base sampler, methods of which are later called.
+   * @param[out] buffer The buffer to hold the adaptation states.
+   * @param[inout] start_gate A latch to gate work to start synchronously across
```
bob-carpenter, line 88: Doesn't this need to be `in,out`, which is what I've used elsewhere.  It looks like they added `inout` as an option but everything I see online says to prefer `in,out` so as not to break other tools that haven't caught up.  

Shouldn't this just be `in`?  We don't actually set the `start_gate` here---it's passed in.  We do change its state, but I don't think that makes it `in, out` by standard conventions.  It's like `rng` in that way.

bob-carpenter (approved):
Thanks.  This looks great.  My only minor question is about `inout` as a form and the particular choice in one instance.  Given that these may be fine as is, I'm just going to mark as approved.

include/walnuts/adapt.hpp
WardBrian: Hm, my editor was refusing to highlight `in,out`, but according to the [doc](https://www.doxygen.nl/manual/commands.html) any of `in,out`, `in out`, `inout`, `out in`, ... are equivalent

PR #83, opened 2026-07-10
Init stepsize

bob-carpenter:
Adds `adapt_step_build(rng, logp_grad)` as alternative to simple `.build()` for `InitConfigBuilder`.  It is the same general idea as the one in NUTS/Stan, but targeting a slightly higher initial acceptance rate and using finer grained stepping.

I also updated `examples/walnuts_api` to use this with a way-too-high starting step size to make sure it's really working.

### DIFF
--- a/include/walnuts/config.hpp
+++ b/include/walnuts/config.hpp
@@ -465,52 +465,13 @@ class InitConfigBuilder {
   template <std::uniform_random_bit_generator RNG, LogpGrad F>
   InitConfig adapt_step_build(RNG& rng, const F& logp_grad) {
     for (std::size_t c = 0; c < num_chains_; ++c) {
-      step_sizes_[c] = adapt_step(rng, logp_grad, positions_[c],
-				  masses_[c], step_sizes_[c]);
+      step_sizes_[c] = detail::adapt_step(rng, logp_grad, positions_[c],
+					  masses_[c], step_sizes_[c], dims_);
     }
     return build();
   }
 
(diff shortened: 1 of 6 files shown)


include/walnuts/config.hpp
```diff
@@ -451,7 +453,64 @@ class InitConfigBuilder {
...
+   * @tparam F Type of the log density and gradient function.
+   * @param[in] rng The base random number generator.
+   * @param[in] logp_grad The log density and gradient function.
+   * @param theta The initial position.
+   * @param M The diagonal of the diagonal mass matrix.
+   * @param step The initial step size guess.
+   * @return The adapted step size.
+   */
+  template <std::uniform_random_bit_generator RNG, LogpGrad F>
+  double adapt_step(RNG& rng, const F& logp_grad,
```
WardBrian: Minor, but I would make this a free function (maybe in util.hpp) rather than a method, since it looks like it doesn't use any other members than the ones passed in?

WardBrian (approved):
Looks good; one small comment on organization

include/walnuts/config.hpp
bob-carpenter: It used `dims_`, but I just added that as another arg and moved the function to `util.hpp`.   I would have left it, or at most converted to a static function, but I can see the argument either way.  It's not really a standalone, so I added a note to the doc saying that it doesn't validate its arguments because it's an internal utility function.

PR #85, opened 2026-07-27
Add callback for error handling in NoExceptLogpGrad

WardBrian:
This adds `void on_logp_exception(const Eigen::VectorXd& position, const std::exception& exn)` to the handler concepts, resolving a TODO for some sort of logging in `NoExceptLogpGrad`

I also moved `handlers.hpp` to `examples/` -- that's the only place it is used, and I don't think we should be providing such things in the public API

### DIFF
--- a/examples/examples.cpp
+++ b/examples/examples.cpp
@@ -4,6 +4,7 @@
 #include <random>
 
 #include <walnuts.hpp>
+#include "handlers.hpp"
 
 static double total_time = 0.0;
 static std::size_t count = 0;
@@ -113,7 +114,7 @@ static void run_adaptive_walnuts(F& target_logp_grad) {
   std::cout << "Warmup configuration:\n" << warmup_cfg << std::endl;
   std::cout << "Sampling configuration:\n" << sampling_cfg << std::endl;
 
-  walnuts::ChainStore handler;
+  ChainStore handler;
   walnuts::AdaptiveWalnuts adapt(rng, handler, target_logp_grad,
                                  init_cfg.init_chain_config(0), warmup_cfg,
                                  sampling_cfg);
--- a/examples/handlers.hpp
+++ b/examples/handlers.hpp
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <fstream>
+#include <iostream>
 #include <stdexcept>
 #include <string>
 #include <type_traits>
@@ -13,9 +14,9 @@
 
 #include <Eigen/Dense>
 
-#include "walnuts/validate.hpp"
+#include <walnuts/validate.hpp>
 
-namespace walnuts::detail {
+namespace detail {
 
 /**
  * @brief Internal flag with value `true` if C++ received `SIGINT`.
@@ -53,9 +54,23 @@ static void write_vector(std::ostream& os, const Eigen::VectorXd& v) {
   os.write(data, size);
 }
 
-}  // namespace walnuts::detail
+/**
+ * @brief Validate that the specified stream is open.
+ *
+ * @tparam S The type of the stream.
+ * @param[in] s The stream.
+ * @param[in] name The name of the stream.
+ * @throw invalid_argument If the stream is not open.
+ */
+template <typename S>
+inline void validate_open(const S& s, const std::string& name) {
+  if (s.is_open()) {
+    return;
+  }
+  throw std::invalid_argument("could not open stream from: " + name);
+}
 
-namespace walnuts {
+}  // namespace detail
 
 /**
  * @brief An interrupt callback for C++.
@@ -157,11 +172,10 @@ class ChainStore {
     lps_.push_back(lp);
   }
 
-  /**
-   * @brief Handle an event to stop sampling.
-   */
-  void on_stop() {
-    // TODO: add stop semaphore, catch interrupts
+  void on_logp_exception(const Eigen::VectorXd& position,
+                         const std::exception& exn) const noexcept {
+    std::cout << "Logp failed with exception " << exn.what() << " at "
+              << position.transpose() << "\n";
   }
 
   /**
@@ -444,5 +458,3 @@ static void write_sample(const std::string& file_name,
   detail::validate_open(os, file_name);
   write_sample(os, handlers, include_warmup);
 }
-
-}  // namespace walnuts
--- a/examples/stan_cli.cpp
+++ b/examples/stan_cli.cpp
@@ -88,6 +88,12 @@ class StanHandler {
   void on_warmup_complete(double step_size,
                           const Eigen::VectorXd& diag_inv_mass) {}
 
+  void on_logp_exception(const Eigen::VectorXd& position,
+                         const std::exception& exn) const noexcept {
+    std::cout << "Logp failed with exception " << exn.what() << " at "
+              << position.transpose() << "\n";
+  }
+
   void summarize() {
     auto names = model_.param_names();
     ::summarize(names, draws_);
--- a/include/walnuts/concepts.hpp
+++ b/include/walnuts/concepts.hpp
@@ -99,19 +99,6 @@ concept Sampler = requires(S& s, const S& cs) {
 };
 
 /**
- * @brief Concept for a stream that reports whether it is open.
- *
- * A type `S` satisfies `OpenableStream` if `s.is_open()` is callable on
- * a const instance and returns a value convertible to `bool`. This
- * matches the interface of standard file streams (`std::ifstream`,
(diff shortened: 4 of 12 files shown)


examples/examples.cpp
```diff
@@ -113,7 +114,7 @@ static void run_adaptive_walnuts(F& target_logp_grad) {
   std::cout << "Warmup configuration:\n" << warmup_cfg << std::endl;
   std::cout << "Sampling configuration:\n" << sampling_cfg << std::endl;
 
-  walnuts::ChainStore handler;
+  ChainStore handler;
```
bob-carpenter, line 117: Presumably if line 117 need `walnuts::`, neither does line 118.  Now that it's all just one namespace, we should just search and remove `walnuts::` wherever it shows up in the API code.

examples/handlers.hpp
```diff
@@ -6,16 +6,17 @@
...
+#include <iostream>
 #include <stdexcept>
 #include <string>
 #include <type_traits>
 #include <vector>
 
 #include <Eigen/Dense>
 
-#include "walnuts/validate.hpp"
+#include <walnuts/validate.hpp>
```
bob-carpenter, line 17: I think this convention may be inconsistent throughout with quotes used in a lot of places for internal includes.  This should be made consistent globally, maybe in another PR.

examples/handlers.hpp
```diff
@@ -157,11 +172,10 @@ class ChainStore {
     lps_.push_back(lp);
   }
 
-  /**
-   * @brief Handle an event to stop sampling.
-   */
-  void on_stop() {
-    // TODO: add stop semaphore, catch interrupts
+  void on_logp_exception(const Eigen::VectorXd& position,
```
bob-carpenter, line 175: The `on_stop` that was removed was, I believe, meant to be a placeholder suggested by @SteveBronder to deal with external interrupts.  I was never sure how to actually deal with those, though.

examples/handlers.hpp
```diff
@@ -157,11 +172,10 @@ class ChainStore {
...
+                         const std::exception& exn) const noexcept {
+    std::cout << "Logp failed with exception " << exn.what() << " at "
+              << position.transpose() << "\n";
```
bob-carpenter, line 178: This output might be huge in high dimensions.  Should we truncate to the first few dimensions?  e.g., `` and maybe add 

```cpp
    << position.head(std::min(50, position.size())).transpose()
```

and ideally a little message after that that says how many elements were elided, though that's a pain to write inline, so I think it'll need a support function.

examples/stan_cli.cpp
```diff
@@ -88,6 +88,12 @@ class StanHandler {
   void on_warmup_complete(double step_size,
                           const Eigen::VectorXd& diag_inv_mass) {}
 
+  void on_logp_exception(const Eigen::VectorXd& position,
```
bob-carpenter, line 91: This is a straight cut-and-paste.  It makes me want to use the CRTP to form a class without virtual functions to deal with this, but that's probably overkill.  If I see another one, I'll suggest pulling out into a shared function.

include/walnuts/concepts.hpp
```diff
@@ -201,17 +188,31 @@ concept InterruptCallback = requires(H& h, const H& ch, double r_hat) {
   { h.throw_if_interrupted() } -> std::same_as<void>;
 };
 
+/**
```
bob-carpenter, line 191: Is there a reason to treat this separately and not just roll it into `SampleHandler`?

include/walnuts/util.hpp
```diff
@@ -240,17 +240,16 @@ inline double logp_momentum(const Eigen::VectorXd& rho,
...
+                      const Eigen::VectorXd& rho, const Eigen::VectorXd& inv_M,
+                      double step) {
   Eigen::VectorXd grad;
   double logp;
   logp_grad(theta, logp, grad);
   logp += detail::logp_momentum(rho, inv_M);
   Eigen::VectorXd rho_star = rho + 0.5 * step * grad;
-  Eigen::VectorXd theta_star = theta + step * (inv_M.array() * rho_star.array()).matrix();
+  Eigen::VectorXd theta_star =
+      theta + step * (inv_M.array() * rho_star.array()).matrix();
```
bob-carpenter, line 252: I have a very strong preference for the `=` sign to go on the next line to make the code easy to scan down the left side.  Is this placement of `=` the result of Clang format?

include/walnuts/util.hpp
```diff
@@ -283,33 +282,33 @@ double leapfrog_error(const F& logp_grad,
...
+template <std::uniform_random_bit_generator RNG, LogpGrad F>
+double adapt_step(RNG& rng, const F& logp_grad, const Eigen::VectorXd& theta,
+                  const Eigen::VectorXd& M, double step, std::size_t D) {
+  detail::Random<RNG> rand(rng);
+  Eigen::VectorXd inv_M = M.array().inverse().matrix();
+  Eigen::VectorXd rho =
+      (rand.standard_normal(static_cast<Eigen::Index>(D)).array() *
+       M.array().sqrt())
+          .matrix();
+  while (detail::leapfrog_error(logp_grad, theta, rho, inv_M, step) >
```
bob-carpenter, line 294: Same with this operator---I'd rather see `>` on the next line.

include/walnuts/util.hpp
```diff
@@ -336,16 +337,17 @@ class NoExceptLogpGrad {
                   Eigen::VectorXd& grad) const noexcept {
     try {
       logp_grad_.get()(x, logp, grad);
-    } catch (...) {
+    } catch (const std::exception& e) {
+      handler_.get().on_logp_exception(x, e);
```
bob-carpenter, line 341: The meat of the PR here looks good.

tests/test_util.hpp
```diff
@@ -6,9 +6,10 @@
 #include <Eigen/Dense>
 
 struct ThrowingLogpGrad {
+  static constexpr char const* const ERRORMSG = "logp_grad failed";
```
bob-carpenter: Is the purpose of this factoring just to make `ERRORMSG` available for testing?  

I'd prefer `ERROR_MSG` but no big deal either way if there's some convention here to not include separators.

tests/test_util.hpp
```diff
@@ -20,6 +21,17 @@ struct GoodLogpGrad {
   }
 };
 
+struct TattleTaleHandler {
```
bob-carpenter, line 24: Love the name.

bob-carpenter (changes requested):
The content here looks good, but I had a bunch of superficial comments, none of which are critical, so feel free to push back.

include/walnuts/util.hpp
WardBrian: Yes, my editor runs clang format on save which sometimes leads to these spurious changes

include/walnuts/util.hpp
WardBrian: Also just the result of clang format

include/walnuts/concepts.hpp
WardBrian: Mainly so that `NoExceptLogpGrad` can have as narrowly-scoped as possible of a requirement

examples/handlers.hpp
WardBrian: It's just for the example code, which is hard coded to 100 dimensions right now (and, coincidentally, never throws), so I'm not sure how much effort it is worth. I am happy to do something more if you'd like

examples/handlers.hpp
WardBrian: I think that was a misunderstanding, and the `InterruptCallback` we have still is what was actually being requested back then

examples/handlers.hpp
WardBrian: It's probably worth cleaning up, but not crucial

examples/examples.cpp
WardBrian: Line 118 still does. I removed the example handlers from the walnuts namespace specifically to make it clear that they were not part of the general API

tests/test_util.hpp
WardBrian: Yep, just to have a good comparison point. I've changed the name

examples/handlers.hpp
bob-carpenter: Oh, I didn't catch that this was for an example.  In which case, don't bother.  If we build something into the API, then we are going to need to truncate output.

examples/handlers.hpp
bob-carpenter: Let me revise my comment to "thanks".

PR #90, opened 2026-08-06
Warn when a seed is specified but adaptive stopping is still enabled

WardBrian:


python/src/walnutpie/pyfunc.py
```diff
@@ -181,7 +181,11 @@ def walnuts_pyfunc(
         else:
             num_params = init_shape[0]
 
-    seed = prepare_seed(seed)
+    is_adaptive = num_chains > 1 and (
```
bob-carpenter: `is_adaptive` is a confusing name here given there's multiple adaptations in play.  Can we call this something like `is_monitored`?  

And how often do we run if `is_monitored` is `False`, but `min_warmup_iter != max_warup_iter`?  Do we just take the minimum?

python/src/walnutpie/__init__.py
```diff
@@ -24,4 +24,4 @@ def walnuts_stan(*args, **kwargs):
     "standard_deviation",
     "Summarizer",
 ]
-__version__ = "0.0.1"
+__version__ = "0.0.2"
```
bob-carpenter, line 27: Isn't this going to need a change in doc somewhere?

python/src/walnutpie/stan.py
```diff
@@ -304,7 +306,11 @@ def walnuts_stan(
             "BridgeStan model must be compiled with STAN_THREADS for use with walnuts"
         )
 
-    seed = prepare_seed(seed)
+    is_adaptive = num_chains > 1 and (
```
bob-carpenter: I hate this kind of duplicated behavior.  Can we pull this out into a function that's used by both `stan.py` and `pyfunc.py`?  Usually I wouldn't say that with only two uses, but this is something I think we want to keep in synch and I won't be obvious there's another one editing either one.

python/src/walnutpie/util.py
```diff
@@ -9,8 +10,18 @@ def rand_u32():
...
 
 
-def prepare_seed(seed: Optional[int]) -> int:
-    return seed if seed is not None else rand_u32()
+def prepare_seed(seed: Optional[int], is_adaptive: bool) -> int:
+    if seed is not None:
+        if is_adaptive:
+            warnings.warn(
+                "Setting 'seed' without also disabling adaptive stopping "
+                "(by setting min and max number of iterations for warmup and sampling) "
```
bob-carpenter: Is this line supposed to say by setting them to the same value?  Because min and max always get at least a default value.

bob-carpenter (changes requested):

python/src/walnutpie/__init__.py
WardBrian: I don't think we actually specify a version in the docs anywhere at the moment, and if we did it would probably only be relevant when this gets actually released

python/src/walnutpie/stan.py
WardBrian: I considered doing it as part of `prepare_seed`, but it felt odd to me that a function for the seed would take in five other seemingly unrelated arguments. If you feel otherwise I'm happy to organize it differently

python/src/walnutpie/pyfunc.py
WardBrian: > And how often do we run if is_monitored is False, but min_warmup_iter != max_warup_iter? Do we just take the minimum?

What do you mean by this?

python/src/walnutpie/__init__.py
bob-carpenter: Sorry, my comment was confusing. I only put it here because this was the top of the file.  It seems like changing behavior should be documented somewhere.  I don't care about the version number as much

python/src/walnutpie/stan.py
bob-carpenter: I was just suggesting writing an `is_adaptive()` that took in the relevant arguments.  Then you can just make that the second argument without the intermediate, or you could keep the `is_adaptive` variable.

python/src/walnutpie/__init__.py
WardBrian: Ah, this is already documented at https://flatironinstitute.github.io/walnutpie/latest/py.html#note-on-reproducibility

The warning is just to serve as a reminder/catch the folks who don't scroll that far

bob-carpenter (approved):
Thanks for factoring out an adaptivity function.

PR #98, opened 2026-08-25
Fix row-major confusion when passing draws to summary functions

### LINKED ISSUE #97: Unexpected `r_hat`, `ess`, `mcse` values for draws with more than one column
Apologies up front for the length of this, but I wanted to be thorough in case I'm missing anything.

I'm working on support for chains of unequal length in ArviZ (arviz-devs/arviz-base#219). Since `walnutpie` already defines [R-hat](https://github.com/flatironinstitute/walnutpie/blob/6162d881a38c4da01fd82ff6e2a1fc582c303009/include/walnutpie/summary.hpp#L594), [ESS](https://github.com/flatironinstitute/walnutpie/blob/6162d881a38c4da01fd82ff6e2a1fc582c303009/include/walnutpie/summary.hpp#L664) and [MCSE](https://github.com/flatironinstitute/walnutpie/blob/6162d881a38c4da01fd82ff6e2a1fc582c303009/include/walnutpie/summary.hpp#L765) for ragged chains in `summary.hpp`, I was messing around with those functions in `numpy` as the reference for what `arviz-stats` will eventually need. I ran into an issue (I think it's an issue at least) with models that have multiple parameters. I wanted to open this to see if I'm missing something.

Passing draws with more than one parameter to `r_hat`, `ess` or `mcse` gives different numbers than passing the same columns one at a time. So consider where we have a second parameter that's just the first one times ten, so R-hat should come out the same for both.

```python
import numpy as np
import walnutpie as wp

chain0 = np.array([[1., 10.], [2., 20.], [3., 30.], [4., 40.]])
chain1 = np.array([[5., 50.], [6., 60.], [7., 70.], [8., 80.]])

wp.r_hat([chain0, chain1])                   # [1.133399 1.025218]
wp.r_hat([chain0[:, :1], chain1[:, :1]])     # [2.408319]
wp.r_hat([chain0[:, 1:], chain1[:, 1:]])     # [2.408319]
```

It goes the other way too. Here the first parameter mixes fine and the second one sits at a different constant in every chain, so its R-hat should be huge, but together it looks fine and the first one gets flagged instead.

```python
rng = np.random.default_rng(0)
chains = [
    np.column_stack([rng.normal(size=200), 10.0 * k + 1e-3 * rng.normal(size=200)])
    for k in range(4)
]

wp.r_hat(chains)                                                        # [1.279916 1.026099]
np.array([wp.r_hat([c[:, [j]] for c in chains])[0] for j in range(2)])  # [1.003804e+00 1.299179e+04]
```

I also used a small PyMC model to get some draws to test a little more thoroughly.

```python
import threading

import arviz_base as azb
import numpy as np
import pymc as pm
import walnutpie as wp

eight = azb.load_arviz_data("centered_eight")
y = eight["observed_data"]["obs"].values
sigma = eight["constant_data"]["sigma"].values

with pm.Model() as model:
    mu = pm.Normal("mu", 0, 5)
    tau = pm.HalfCauchy("tau", 5)
    theta = pm.Normal("theta", mu, tau, shape=8)
    pm.Normal("y", theta, sigma, observed=y)

value_grad = model.logp_dlogp_function(ravel_inputs=True)
value_grad.set_extra_values({})
lock = threading.Lock()  # walnutpie calls logp from one thread per chain

def logp(x):
    with lock:
        return value_grad(x)
```

I made all four chains the same length so I can check the summary stats against ArviZ without unequal chain lengths being another difference to account for.

```python
chains = wp.walnuts_pyfunc(
    logp, num_params=10, num_chains=4, seed=315,
    min_warmup_iter=1000, max_warmup_iter=1000,
    min_sampling_iter=2000, max_sampling_iter=2000,
)
```

I calculated R-hat and ESS with all ten parameters at once, one parameter at a time, and with `arviz-stats` on the same draws as an outside reference. `method="identity"` uses the same R-hat and ESS estimators as `walnutpie`, except that `walnutpie` replaces the `(N - 1) / N` factor with 1 (the [`r_hat` docstring](https://github.com/flatironinstitute/walnutpie/blob/6162d881a38c4da01fd82ff6e2a1fc582c303009/python/src/walnutpie/summary.py#L175-L178) mentions this for R-hat, and the same factor sits inside ArviZ's within-chain variance for ESS), so I add `1/N` back for the R-hat comparison.

```python
import arviz_stats as azs

rhat_together, ess_together = wp.r_hat(chains), wp.ess(chains)

rhat_separately = np.array([wp.r_hat([c[:, [j]] for c in chains])[0] for j in range(10)])
ess_separately = np.array([wp.ess([c[:, [j]] for c in chains])[0] for j in range(10)])

idata = azb.from_dict({"posterior": {"x": np.stack(chains)}}, dims={"x": ["param"]})
rhat_arviz = np.sqrt(azs.rhat(idata, method="identity")["x"].values ** 2 + 1 / 2000)
ess_arviz = azs.ess(idata, method="identity")["x"].values
```

```text
              rhat        rhat       rhat       ess         ess        ess
          together  separately      arviz  together  separately      arviz
mu          1.0059      1.0054     1.0054     250.1       500.7       501.9
log_tau     1.0124      1.0250     1.0250     416.4       112.4       112.6
theta[0]    1.0065      1.0013     1.0013     394.8       818.3       821.9
theta[1]    1.0169      1.0021     1.0021     231.5       987.0       991.6
theta[2]    1.0023      1.0020     1.0020     241.9      1056.4      1062.2
theta[3]    1.0202      1.0030     1.0030     112.9       886.6       891.6
theta[4]    1.0072      1.0038     1.0038     312.7       773.7       777.5
theta[5]    1.0121      1.0019     1.0019     293.5       835.4       840.0
theta[6]    1.0355      1.0036     1.0036      74.0       681.0       683.2
theta[7]    1.0134      1.0015     1.0015     133.0      1037.3      1045.1
```

I expected all three to agree. The one-at-a-time results match ArviZ to rounding for R-hat and to within about a percent for ESS (ArviZ keeps the `(N - 1) / N` factor on the within-chain variance inside its ESS calculation and `walnutpie` doesn't, which is where the small ESS gap comes from), but passing everything together gives pretty different numbers. As you can see, for `theta[6]`, R-hat goes from about 1.004 to 1.036 and ESS goes from about 680 to 74. `log_tau` goes the other way, from about 1.025 one at a time to 1.012 together, with its ESS going from about 110 to 420. I get the same kind of difference with `mcse` as well.

I tried it with adaptive stopping on too, so the chains come out with different lengths (same call with `min_warmup_iter=300, max_warmup_iter=1500, min_sampling_iter=300, max_sampling_iter=3000`). In one run the chains stopped at 3000, 2393, 3000 and 3000 draws. One parameter at a time, every R-hat was at or below 1.009. With all ten parameters together, five were above 1.01. `theta[4]`, for example, came out at 1.022 with an ESS of 71, compared with about 1.002 and 750 when I ran it separately.

Just wondering if this is expected behavior? I don't see why they should be different, but maybe I'm misunderstanding something.

WardBrian:
Fixes #97

### DIFF
--- a/python/src/walnutpie/summary.py
+++ b/python/src/walnutpie/summary.py
@@ -27,8 +27,8 @@ class Summarizer:
 
         if hasattr(draws[0], "parameters"):  # StanOutputBase
             draws = [c.data for c in draws]
-        self._stacked = np.concat(draws)
-        self._num_draws, self._num_params = self._stacked.shape
+        self._stacked = np.ascontiguousarray(np.concat(draws).transpose())
+        self._num_params, self._num_draws = self._stacked.shape
 
         self._lengths = np.array([c.shape[0] for c in draws], dtype=np.int32)
         self._num_chains = len(draws)
--- a/python/tests/test_summary.py
+++ b/python/tests/test_summary.py
@@ -3,7 +3,7 @@ import walnutpie as wp
 
 
 def test_chain_ordering():
-
+    # test that there is no difference between calculating stats per-parameter versus all at once
     chains = [
         np.array([[1.0, 10.0], [2.0, 20.0], [3.0, 30.0], [4.0, 40.0]]),
         np.array([[5.0, 50.0], [6.0, 60.0], [7.0, 70.0], [8.0, 80.0]]),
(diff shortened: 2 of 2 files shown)


WardBrian:
@bob-carpenter once this is merged I'd like to do a 0.2 release on PyPI

bob-carpenter:
A 0.2 release is fine with me.  Thanks for patching.

PR #103, opened 2026-09-08
Validate both chain-view indices before accessing storage

sims1253:
`MarkovChainsUnified::chain_view()` promises `std::out_of_range` for an invalid chain index, but its two arguments mix checked and unchecked indexing, maybe just a typo?

C++ does not require the checked argument to run first. GCC 16 reproduces an out-of-bounds read before the exception. With libstdc++ assertions enabled, the existing bounds test aborts.


WardBrian (approved):
Thanks

PR #104, opened 2026-09-09
Treat non-finite acceptance statistics as rejection before adaptation

sims1253:
Nonfinite Hamiltonian differences can produce a NaN acceptance statistic and poison step-size adaptation.

### DIFF
--- a/tests/walnuts_test.cpp
+++ b/tests/walnuts_test.cpp
@@ -1,0 +1,98 @@
+#include <gtest/gtest.h>
+#include <walnutpie/adam.hpp>
+#include <walnutpie/walnuts.hpp>
+
+#include <cmath>
+#include <limits>
+#include <vector>
+
+namespace {
+using walnutpie::detail::Adam;
+using walnutpie::detail::Direction;
+using walnutpie::detail::SpanW;
+
+struct RecordingAdapter {
+  std::vector<double> observed;
+  void operator()(double alpha) { observed.push_back(alpha); }
+  double step_size() const { return 1.0; }
+};
+
+template <Direction D, typename F, typename A>
+bool take_macro_step(const F& logp_grad, A& adapter,
+                     std::size_t max_halvings = 3) {
+  auto span = SpanW::from_initial_point(Eigen::VectorXd::Zero(1),
+                                        Eigen::VectorXd::Ones(1),
+                                        Eigen::VectorXd::Zero(1), 0.0, -0.5);
+  Eigen::VectorXd theta(1), rho(1), grad(1);
+  double logp_pos, logp;
+  return walnutpie::detail::macro_step<D>(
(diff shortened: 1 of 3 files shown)


tests/walnuts_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+TEST(MacroStepAdaptation, ConservedEnergyKeepsUnitAcceptance) {
+  auto target = [](const Eigen::VectorXd&, double& lp, Eigen::VectorXd& grad) {
+    lp = 0.0;
+    grad.setZero();
+  };
+  RecordingAdapter adapter;
+  EXPECT_TRUE(take_macro_step<Direction::Forward>(target, adapter));
+  EXPECT_EQ(adapter.observed, std::vector<double>{1.0});
```
bob-carpenter, line 98: Nice---I hadn't realized you could just compare standard vectors, but it makes sense given how their `==` operator is implemented as having the same size and matching values.

tests/walnuts_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+  EXPECT_FALSE(take_macro_step<Direction::Forward>(target, adapter));
+  EXPECT_EQ(adapter.observed, std::vector<double>{0.0});
+}
+
+TEST(MacroStepAdaptation, AdamMatchesRejectionAndRemainsUsable) {
+  auto target = [](const Eigen::VectorXd&, double& lp, Eigen::VectorXd& grad) {
+    lp = std::numeric_limits<double>::quiet_NaN();
+    grad.setZero();
+  };
+  Adam actual(1.0, 0.8, 0.05, 0.9, 0.999, 1e-8, 0.5);
```
bob-carpenter, line 67: The constants don't matter for this test, but this is an amusing mix of the Walnutpie defaults (e.g., 0.9, 0.05) and standard Adam defaults (e.g., 0.999, 1e-8).

tests/walnuts_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+                                        Eigen::VectorXd::Ones(1),
+                                        Eigen::VectorXd::Zero(1), 0.0, -0.5);
+  Eigen::VectorXd theta(1), rho(1), grad(1);
+  double logp_pos, logp;
+  return walnutpie::detail::macro_step<D>(
+      logp_grad, Eigen::VectorXd::Ones(1), 1.0, max_halvings, 1, 1.0, span,
+      theta, rho, grad, logp_pos, logp, adapter);
+}
+
+TEST(MacroStepAdaptation, NonfiniteLogDensityIsZeroAcceptanceInBothDirections) {
```
bob-carpenter, line 35: I like this test pattern over some non-finite `double` values.

bob-carpenter (commented):
This looks good.  I'm going to make sure the tests actually run, then merge.

WardBrian (approved):

PR #105, opened 2026-09-09
Add opt-in deterministic endpoint reuse

sims1253:
Reuse the selected density and gradient instead of evaluating them again at the next transition.

Add opt-in `EndpointReuse::Deterministic` to `WalnutsSampler` and `AdaptiveWalnuts`. The default stays disabled as targets may read a mutable external state.
Caches own finite values, transfer across freeze, and support explicit invalidation. Existing constructor calls and the uncached transition API remain valid.

Three repetitions × four chains × 1,000 warmup + 1,000 draws, with paired seeds and identical model binaries:
| Model | Median enabled/disabled ESS/s | Rep range |
|---|---:|---:|
| dogs_hierarchical | 1.190× | 1.187-1.194× |
| gp_regr | 1.188× | 1.170-1.211× |

ESS/s includes warmup and sampling. Each chain saves 1,999 evaluations (warmup + draws - 1). 

Opt-in requires a stable, position-only target. Invalidate after target changes. Failed endpoints are retried.

### DIFF
--- a/docs/cpp.rst
+++ b/docs/cpp.rst
@@ -55,26 +55,14 @@ The following concepts describe the types expected by `walnutpie`.
 .. doxygenconcept:: walnutpie::GlobalHandler
 .. doxygenconcept:: walnutpie::InterruptCallback
 
-Endpoint reuse (opt-in)
------------------------
+Endpoint reuse
+--------------
 
-``WalnutsSampler`` and ``AdaptiveWalnuts`` accept a final constructor argument,
-``walnutpie::EndpointReuse::Deterministic``. The default is ``Disabled``.
-For a stable target, reuse avoids one density/gradient evaluation per transition
-following the first. A finite warmup endpoint also transfers to the fixed sampler.
-With zero warmup, the fixed sampler evaluates its first endpoint normally.
-
-Opt in only when the target returns the same results for the same position.
-Target results must not depend on evaluation counts or handler side effects.
-After changing target data between draws, call ``invalidate_endpoint_cache()``
-on every affected sampler. A warmup-complete handler that changes target data
-must invalidate the adaptive sampler before it returns, or the caller must
-invalidate the returned fixed sampler before drawing. Invalidation does not
-make samples from a changing target suitable for inference.
-
-Reuse retains only finite density and gradient values. Failed or nonfinite
-endpoints are reevaluated. Draw and warmup callbacks retain their order and
-frequency; error callbacks report actual thrown evaluations, not cache hits.
-Caches are copied by value; RNG, target, and handler references retain their
-existing ownership rules. The high-level configuration and Python API do not
-enable this option.
+Samplers reuse finite endpoint density/gradient values automatically. Targets
+must return the same results for the same position. After changing target data,
+call ``invalidate_endpoint_cache()`` on each affected sampler, including from
+callbacks that make such changes. A warmup-complete callback must invalidate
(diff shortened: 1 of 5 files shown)


docs/cpp.rst
```diff
@@ -54,3 +54,27 @@ The following concepts describe the types expected by `walnutpie`.
...
+Endpoint reuse (opt-in)
+-----------------------
+
+``WalnutsSampler`` and ``AdaptiveWalnuts`` accept a final constructor argument,
+``walnutpie::EndpointReuse::Deterministic``. The default is ``Disabled``.
+For a stable target, reuse avoids one density/gradient evaluation per transition
+following the first. A finite warmup endpoint also transfers to the fixed sampler.
+With zero warmup, the fixed sampler evaluates its first endpoint normally.
+
+Opt in only when the target returns the same results for the same position.
```
WardBrian: I’m pretty sure this is always an assumption for MCMC, no?

sims1253: I think so, but the current contract allows the external state to change. Which might involuntary I suppose, in which case I could change that and make this the default?

docs/cpp.rst
```diff
@@ -54,3 +54,15 @@ The following concepts describe the types expected by `walnutpie`.
...
+--------------
+
+Samplers reuse finite endpoint density/gradient values automatically. Targets
+must return the same results for the same position. After changing target data,
+call ``invalidate_endpoint_cache()`` on each affected sampler, including from
+callbacks that make such changes. A warmup-complete callback must invalidate
+the adaptive sampler before returning, or invalidate the returned fixed sampler
+before drawing. This does not make changing-target draws valid for inference.
+Failed/nonfinite endpoints are reevaluated. Zero warmup starts with no endpoint;
+otherwise the fixed sampler inherits the adaptive sampler's finite endpoint.
```
WardBrian: I don't think this is necessary documentation. It's largely meaningless

include/walnutpie/adaptive_walnuts.hpp
```diff
@@ -262,14 +267,22 @@ class AdaptiveWalnuts {
...
+    result.endpoint_gradient_ = endpoint_gradient_;
+    result.endpoint_logp_ = endpoint_logp_;
+    result.endpoint_valid_ = endpoint_valid_;
+    return result;
   }
 
+  /** Targets must be stable at each position. Call after target-data changes,
+   * including changes in callbacks, to force the next endpoint evaluation. */
+  void invalidate_endpoint_cache() noexcept { endpoint_valid_ = false; }
+
```
WardBrian: I dislike this kind of API and I'm still not sure why someone would ever need to call it

include/walnutpie/walnuts.hpp
```diff
@@ -562,6 +567,21 @@ inline Eigen::VectorXd transition_w(
   return std::move(span_accum.theta_select_);
 }
 
+/** Uncached entry point retained for existing callers. */
```
WardBrian: We're in version 0.0.3, we should feel free to not do this sort of hedging

WardBrian (changes requested):
I think the LLM is still hedging a bit too much, and the code still isn't very clean. 

In particular, I suspect if you call logp_grad once at the start of the existence of these objects, you can probably remove all the `endpoint_valid_` flags and branching?

WardBrian (approved):
Thanks!

PR #106, opened 2026-09-10
Reject mass_init_count of one or less at the builder

sims1253:
The mass estimator's discount factor is `1 - 1/(mass_init_count + iteration)`. The old finite-positive check admitted values that break downstream:

- `1` zeroes the discount on the first observation. Both variance priors are erased (`weight = 1`, `sum_sq_dev = 0`), and `inv_mass_estimate()` computes `sqrt(0/0)` = NaN. Verified on the real `MassEstimator`: unit mass, position 1, gradient -1, count 1 gives NaN; count 4 gives 1.0.
- below `1` the discount is negative and `OnlineMoments::discount_observe` throws `discount_factor must be in [0, 1]`. In `AdaptWorker` that exception escapes the thread function and terminates the process.

### DIFF
--- a/tests/config_test.cpp
+++ b/tests/config_test.cpp
@@ -617,7 +617,9 @@ TEST(WarmupConfigBuilder, MassInitCountSetsCorrectly) {
 }
 
 TEST(WarmupConfigBuilder, MassInitCountThrowsOnBadValues) {
-  for (auto x : inf_nan_neg_zero()) {
+  std::vector<double> bad = inf_nan_neg_zero();
+  bad.insert(bad.end(), {0.5, 1.0});
+  for (auto x : bad) {
     walnutpie::WarmupConfigBuilder b;
     EXPECT_THROW(b.mass_init_count(x), std::invalid_argument);
   }
--- a/include/walnutpie/config.hpp
+++ b/include/walnutpie/config.hpp
@@ -696,10 +696,10 @@ class WarmupConfigBuilder {
    * @param[in] v The mass matrix estimator initial count.
    * @return This builder for chaining.
    * @throw std::invalid_argument If the initial count is not finite and
-   * positive.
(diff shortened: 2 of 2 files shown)


tests/config_test.cpp
```diff
@@ -978,3 +978,19 @@ TEST(WalnutsConfig, MembersAreIndependent) {
   EXPECT_DOUBLE_EQ(cfg.init().step_size(0), 0.25);
   EXPECT_DOUBLE_EQ(cfg.init().step_size(1), 0.25);
 }
+
+// class WarmupConfigBuilder ********************************************
+
+TEST(WarmupConfigBuilder, MassInitCountMustExceedOne) {
```
bob-carpenter: To try to keep these already sprawling tests from sprawling any further, these should be rolled into existing tests for WarmupConigBuilder (line 537), not inserted at the end.  Then it'll be clear that the NO_THROW test being introduced here is redundant.  The "new" illegal values can just be added.

bob-carpenter (changes requested):
Thanks!  This is the right fix and right doc for the fix.  I'm only asking that the tests are merged with the other tests of this function.  If you'd rather not, I'm happy moving the tests myself.

tests/config_test.cpp
```diff
@@ -978,3 +981,5 @@ TEST(WalnutsConfig, MembersAreIndependent) {
   EXPECT_DOUBLE_EQ(cfg.init().step_size(0), 0.25);
   EXPECT_DOUBLE_EQ(cfg.init().step_size(1), 0.25);
 }
+
```
bob-carpenter: Thanks---just need to delete the residual comment here and we're good to go.

bob-carpenter (approved):
Thanks much for the corrections!

PR #107, opened 2026-09-10
Report completed doublings from transition_w

sims1253:
`transition_w`'s `depth` output counts attempted doublings, and the three loop exits disagree with the trajectory it actually built:

- U-turn between spans: the doubling landed; `2^depth` states. Correct.
- Failed expansion (macro step fails or a u-turn inside the new span): the trajectory keeps `2^(depth-1)` states, but `depth` names the attempt.
- Budget exhaustion: `depth == max_depth + 1` while the trajectory holds `2^max_depth` states.

The only in-tree consumer is `min_micro_estimator_.observe(1 << depth)` in `AdaptiveWalnuts::operator()`, documented as receiving the micro steps taken. On every capped or aborted warmup draw it records twice the actual states, biasing `min_micro_steps()` upward: longer macro steps than `max_macro_steps_target` calls for, paid in `logp_grad` calls. 

The fix counts the doublings that land and reports that value

### DIFF
--- a/include/walnutpie/walnuts.hpp
+++ b/include/walnutpie/walnuts.hpp
@@ -512,7 +512,8 @@ static std::optional<SpanW> build_span(Random<RNG>& rng, const F& logp_grad,
  * @param[in] min_micro_steps The minimum number of micro steps per macro step.
  * @param[in] max_error The maximum difference in Hamiltonians.
  * @param[in] theta The current state.
- * @param[out] depth The tree depth used by the transition.
+ * @param[out] depth The number of doublings included in the final
+ * trajectory, which holds `2^depth` states.
  * @param[in,out] theta_grad The gradient of the log density at `theta` on
  * input; the gradient at the selected state on output.
  * @param[in,out] logp_pos_select The log density of `theta` on input; the
@@ -534,12 +535,12 @@ inline Eigen::VectorXd transition_w(
   auto span_accum = SpanW::from_initial_point(std::move(theta), std::move(rho),
                                               std::move(theta_grad),
                                               logp_pos_select, logp_joint);
-  for (depth = 1; depth <= max_depth; ++depth) {
-    // helper to turn runtime direction into compile-time template enum
+  depth = 0;
+  while (depth < max_depth) {
     auto expand_in_direction = [&](auto direction) -> bool {
       constexpr Direction D = direction;
       auto maybe_next_span = build_span<D>(
-          rand, logp_grad, inv_mass, step, depth - 1, max_step_halvings,
+          rand, logp_grad, inv_mass, step, depth, max_step_halvings,
           min_micro_steps, max_error, span_accum, step_size_adapter);
       if (!maybe_next_span) {
         return true;
@@ -547,6 +548,7 @@ inline Eigen::VectorXd transition_w(
       bool combined_uturn = uturn<D>(span_accum, *maybe_next_span, inv_mass);
       span_accum = combine<Update::Metropolis, D>(rand, std::move(span_accum),
                                                   std::move(*maybe_next_span));
+      ++depth;
       return combined_uturn;
     };
 
--- a/tests/walnuts_test.cpp
+++ b/tests/walnuts_test.cpp
@@ -4,6 +4,7 @@
 
 #include <cmath>
 #include <limits>
+#include <random>
 #include <vector>
 
 namespace {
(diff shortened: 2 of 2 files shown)


include/walnutpie/walnuts.hpp
```diff
@@ -512,7 +512,8 @@ static std::optional<SpanW> build_span(Random<RNG>& rng, const F& logp_grad,
  * @param[in] min_micro_steps The minimum number of micro steps per macro step.
  * @param[in] max_error The maximum difference in Hamiltonians.
  * @param[in] theta The current state.
- * @param[out] depth The tree depth used by the transition.
+ * @param[out] depth The number of doublings in
```
bob-carpenter: Which trajectory?  Is this the number of doublings attempted, or the ones included in the final set of candidates?

include/walnutpie/walnuts.hpp
```diff
@@ -534,19 +535,22 @@ inline Eigen::VectorXd transition_w(
   auto span_accum = SpanW::from_initial_point(std::move(theta), std::move(rho),
                                               std::move(theta_grad),
                                               logp_pos_select, logp_joint);
-  for (depth = 1; depth <= max_depth; ++depth) {
+  // Doublings that have landed in the accumulated span.
```
bob-carpenter: Drop all the comments---this is just AI-generated over-commenting what the code clearly does.  It's not clear what "landed" means here.  

Rather than a comment in a location like this, it's better to rename the variable.  Although I don't know what "landed" means, you can change the name from `completed` to `non_rejected_doublings` or something more explicit like that.

include/walnutpie/walnuts.hpp
```diff
@@ -534,19 +535,22 @@ inline Eigen::VectorXd transition_w(
...
+  std::size_t completed = 0;
+  for (std::size_t attempted = 1; attempted <= max_depth; ++attempted) {
     // helper to turn runtime direction into compile-time template enum
```
bob-carpenter: if this is what it's doing, name the lambda accordingly rather than including inline comments.

Inline code comments should be thought of as a last resort when it's impossible to write readable and idiomatic code.

include/walnutpie/walnuts.hpp
```diff
@@ -558,6 +562,7 @@ inline Eigen::VectorXd transition_w(
       break;
     }
   }
+  depth = completed;
```
bob-carpenter: Why are there two variables here, `depth` and `completed`?  Can't we just use `depth` in place of `completed` in the code?

tests/walnuts_test.cpp
```diff
@@ -97,4 +98,39 @@ TEST(MacroStepAdaptation, ConservedEnergyKeepsUnitAcceptance) {
...
+TEST(TransitionDepth, ReportsCompletedDoublings) {
+  auto gaussian = [](const Eigen::VectorXd& x, double& lp, Eigen::VectorXd& g) {
+    lp = -0.5 * x.squaredNorm();
+    g = -x;
+  };
+  // A small step doubles until the budget is gone.
+  for (std::size_t max_depth : {std::size_t{1}, std::size_t{3}}) {
+    EXPECT_EQ(transition_depth(gaussian, 0.05, max_depth), max_depth);
+  }
+  // A wall at unit radius fails the first macro step.
```
bob-carpenter: This is an odd way to say this.  I think what you want to say is that a distribution with support on a disk will fail if you try to move outside the disk.  But there's no reason to make this that complicated.  Just have something that always returns -infinity and 0 and it'll work just as well here.  

Also, get rid of the doc, as it'll be transparent what's going on when this is simplified.

bob-carpenter (commented):
Thanks, @sims1253.  This also looks useful.  In this case, I think we may be able to tighten up the code and we can definitely tighten the tests and remove the superfluous doc.  

I hope these changes aren't too onerous---as I've said, I'd be happy to do them myself.

bob-carpenter (changes requested):
I meant that to be an actual review requesting changes, so let's fix.

tests/walnuts_test.cpp
```diff
@@ -97,4 +98,37 @@ TEST(MacroStepAdaptation, ConservedEnergyKeepsUnitAcceptance) {
   EXPECT_TRUE(take_macro_step<Direction::Forward>(target, adapter));
   EXPECT_EQ(adapter.observed, std::vector<double>{1.0});
 }
+// One transition from the origin on a two-dimensional target.
```
bob-carpenter: Thanks for cleaning up.  This looks like a comment that's not attached to anything.

sims1253: Will you stop reviewing my draft slop Bob :D Let me clean up in peace <3

bob-carpenter (approved):
Thanks again for updating.  The loop bounds are better your way and the testing is easy to read.

PR #108, opened 2026-09-10
Take the absolute step deviation from the geometric mean

sims1253:
`controller_loop`'s step-agreement criterion measures only the signed deviation of each chain's step from the geometric mean. A chain sitting below the mean contributes a negative value, which `fmax` discards.

I didn't find any controller test to fold this one into.


bob-carpenter (changes requested):
This is a good change.  We should be looking at absolute error here.  

But I need some help understanding the test, as I can't quite tell what it is intended to test.  I see the stepwise tolerance setup and I see the three step sizes being emplaced as copies.  But then I don't see why the controller loop's going to throw here.  The iteration on the snap instances are all the minimum number of iterations.  So this is a case where maybe a bit of doc would help.  And maybe a case that doesn't throw would help, too.

I also think there's a merge conflict with the last set of test includes.

bob-carpenter:
Thanks for the explanation---that makes sense and adding the simpler test that passes would be helpful.

PR #109, opened 2026-09-11
Fix configuration ownership in AdaptiveWalnuts

sims1253:
`AdaptiveWalnuts` retained a reference to the sampling config. Passing a temporary left a dangling reference after construction.

### DIFF
--- a/include/walnutpie/adaptive_walnuts.hpp
+++ b/include/walnutpie/adaptive_walnuts.hpp
@@ -185,17 +185,12 @@ class AdaptiveWalnuts {
   /**
    * @brief Construct an adaptive Walnuts sampler.
    *
-   * The configuration objects, the base random number generator, and
-   * the log density/gradient function are held by reference. The RNG
-   * changes state every time a random number is generated. The
-   * target depth specifies the expected Nuts tree depth, which is
-   * controlled through the minimum number of micro steps per macro
-   * step and adjusted with a mean estimator to achieve this average.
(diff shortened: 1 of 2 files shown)


bob-carpenter:
> Passing builder temporaries introduced a lifetime error

Thanks for clarifying.  That makes sense and it's an error I could see other people making. 

> With additional template arguments we could make the class store a copy on rvalue inputs and a reference on lvalue inputs

I actually like this perfect forwarding pattern, though I agree it's a bit of a pain to finagle.  Would we then have to change the return type for the builders to be an rvalue reference or do we get that for free?

> the patch copies only SamplingConfig, which currently contains seven scalar fields.

That's obviously going to be fine to copy, but I'd rather not treat the different configurations separately as it feels confusing.

> It also removes the unused AdaptiveWalnuts::warmup_cfg_ member. 

We should at the very least push this fix through.

WardBrian:
> Would we then have to change the return type for the builders to be an rvalue reference or do we get that for free?

The returns would automatically be rvalues if used inline in the constructor

bob-carpenter:
I'm in favor then if we don't need to complicate the builders.  Mainly because I don't like having to write out contracts that people hold references, even if it is relatively natural like in this case.

We could use the perfect forwarding approach with a variant member variable, and that would let us either store a reference wrapper or move the value.  That's ugly at use time, so I'd also be OK with having it move rvalues and copy lvalues, though that's not the usual perfect forwarding.

